PR #236, opened 2025-03-01
release on arm64 natively

shortishly:


shortishly:
only use cross compilation for the multi-arch docker image... release packages should be built on their native architecture

PR #272, opened 2025-04-07
improvement: Panic with concrete failure reason for prometheus listener

zuston:
When trying to start tansu process, I failed to specify the prometheus listener's port and then panic without any reason. 

This pr is to throw concrete failure reason

### DIFF
--- a/tansu-server/src/main.rs
+++ b/tansu-server/src/main.rs
@@ -83,7 +83,9 @@ async fn main() -> Result<()> {
     let mut set = JoinSet::new();
 
     _ = set.spawn(async move {
-        otel::prom::init(prometheus_listener_url).await.unwrap();
+        if let Err(e) = otel::prom::init(prometheus_listener_url).await {
+            panic!("Errors on initializing prometheus listener. error: {}", e);
+        }
     });
 
     let schemas = args.schema_registry.map_or(Ok(None), |schema| {


solace-aross (changes requested):
Thanks for this PR, and for taking the time to report a real problem when you hit it. I'm still new to contributing here and learning my way around the codebase, so I hope this is helpful.

This change targets `tansu-server/src/main.rs`, which no longer exists. That file moved through a couple of renames (`tansu-server` -> `tansu-broker` -> the current `nisshi-*` crates), and separately, the Prometheus listener itself was removed in `6cc3f86` ("Remove Prometheus metrics in favor of OTLP metrics"). GitHub also flags this PR as conflicting against `main`, so it can't be merged as-is.

The encouraging part: the concern behind this PR, an `unwrap()` swallowing an init failure with no useful message, is already solved in the current code. `nisshi-otel::meter_provider()` and `nisshi-broker::otel::init()` both return a typed `Result`, and every call site propagates it with `?` (e.g. `nisshi/src/bin/nisshi.rs:34`, `nisshi-broker/src/broker.rs:663`, `nisshi-proxy/src/lib.rs:327`, `nisshi-generator/src/lib.rs:286`). There's no `unwrap`/`expect`/`panic!` left in that init path.

Given that, there isn't anything left to rebase here. I'd recommend closing this one as superseded rather than reworking it. Really appreciate you flagging the original issue though, that's exactly the kind of instinct that makes error handling better across the codebase.

PR #279, opened 2025-04-14
Switch environment variable for Broker for `tansu cat` commands

Tenzer:
This switches the environment variable that can be used to configure the broker URL for the `tansu cat consume` and `tansu cat produce` commands from `ADVERTISED_LISTENER_URL` (which doesn't really make sense here) to `BROKER_URL`.

---

While working on this, I also noticed the `tansu cat consume --partition-max-bytes` flag, which has a help text of "The partition to consume from" and a default value of `1048576`. This doesn't really make sense to me. I looked through the source code and couldn't really work out what the flag is expected to do (I'm a Rust rookie). However, I thought I would mention it here if that's something that should be corrected at the same time.

### DIFF
--- a/tansu-cli/src/cli/cat.rs
+++ b/tansu-cli/src/cli/cat.rs
@@ -24,7 +24,7 @@ use url::Url;
 pub(super) enum Command {
     #[command(about = "Produce Avro/JSON/Protobuf messages to a topic")]
     Produce {
-        #[arg(long, default_value = DEFAULT_BROKER, env = "ADVERTISED_LISTENER_URL", help = "The URL of the broker to produce messages into")]
+        #[arg(long, default_value = DEFAULT_BROKER, env = "BROKER_URL", help = "The URL of the broker to produce messages into")]
         broker: Url,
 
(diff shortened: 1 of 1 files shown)


shortishly:
I'm not sure about adding another environment variable: `BROKER_URL` should be the same as `ADVERTISED_LISTENER_URL`. Advertised listener is the address used to contact the broker (it is published by the broker in the metadata response, which is used by clients to discover and connect to brokers).  The Apache Kafka clients use "--bootstrap-servers" which then use a metadata request to do broker discovery (to get the advertised listener). We could rename `ADVERTISED_LISTENER_URL` to `BROKER_URL`? However, I think people are familiar with the concept of the advertised listener address (used by Apache Kafka).

Thanks for taking the time to dig through the code on `--partition-max-byte` - it was mostly a todo reminder for me to implement in the fetcher. Could you raise a separate issue for this?

Tenzer:
If the environment variable names are a deliberate choice, then we can close this PR. I assumed it had been copy/pasted from the other commands without taking into account what a descriptive name would be here.

I have created #282 regarding the `--partition-max-bytes` flag.

PR #306, opened 2025-05-26
metadata.date

shortishly:
Add `date` to the `metadata` that is exposed to iceberg and delta lake.


shortishly:
`metadata.year`, `metadata.month` and `metadata.day` added to the existing metadata.timestamp for Protobuffer, Avro and JSON schema.

PR #315, opened 2025-05-27
partition specification for iceberg from topic configuration

### LINKED ISSUE #313: implement `tansu.lake.partition` for iceberg
replicate the functionality from delta lake to use topic configuration `tansu.lake.partition` to partition data into iceberg

shortishly:


tansu-schema-registry/src/lake/berg.rs
```diff
@@ -288,13 +299,64 @@ impl LakeHouse for Iceberg {
...
+                }
+            })
+            .collect::<Vec<_>>()
+    }
+
+    fn partition_fields(&self, schema: &Schema) -> impl Iterator<Item = UnboundPartitionField> {
+        self.as_columns("tansu.lake.partition")
+            .into_iter()
+            .filter_map(|field_name| {
+                schema.field_by_name(&field_name).map(|field| {
```
solace-aross, line 339: `schema.field_by_name(&field_name)` returning `None` here (unmatched/typo'd config value) is silently dropped by `filter_map`, so a misconfigured `tansu.lake.partition` column name results in an unpartitioned table with no diagnostic. Given this is driven by user-supplied topic config, an explicit error (or at minimum a `warn!(?field_name)`) on a name that doesn't resolve against the schema would make misconfiguration visible instead of silently falling back to no partitioning.

solace-aross (commented):
Thanks for this, Peter — I'm still fairly new to contributing here, so please take my notes in that spirit; happy to be corrected on anything I've misread.

I held off on a formal approve/request-changes verdict since this is still marked as a draft, and it looks like it predates a couple of big changes on main: the `tansu-schema-registry` -> `nisshi-schema` rename and the move to compile-time storage feature flags (this crate builds unconditionally today, no `iceberg` feature gate yet on this branch). So it'll need a rebase before it's mergeable regardless of the review outcome.

On the approach itself, `Config::partition_fields` looks like a solid, well-tested first cut for identity partitioning driven by topic config. One thing worth considering before this comes off draft: when a configured `tansu.lake.partition` column name doesn't match any field in the schema (typo, wrong case, renamed column), `filter_map` silently drops it and the table is created unpartitioned with no error or log line. That could be a quiet surprise for an operator who set the config expecting partitioning and gets none. Might be worth an explicit error (or at least a `warn!`) on an unresolved partition column name, rather than a silent drop — happy to help test that if useful.

Great to see partitioning land here either way, this fills a real gap for the iceberg lake path.

PR #332, opened 2025-06-10
proxy - initial use case: batching of produce request

shortishly:
work in progress: proxy for metadata and batch produce requests together

- [x] initial tests for the produce batching process
- [x] skeleton using [rama](https://github.com/plabayo/rama)/tower service trait, which will probably flow into broker 
- [x] when batching ensure that all requests are the same protocol version
- [x] proxied metadata requests replace the broker response with the proxy listener address 
- [ ] ensure that transactions are not batched

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6314,6 +6314,7 @@ dependencies = [
  "dotenv",
  "fake",
  "futures",
+ "governor",
  "iceberg",
  "iceberg-catalog-memory",
  "iceberg-catalog-rest",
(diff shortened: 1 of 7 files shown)


shortishly:
The proxy will collect produce requests, which are batched into a single produce request. The single batched request is sent to the broker, the broker writes the request to storage,  the proxy waits for the reply from the broker. The proxy will only reply to the original client requests at that point - splitting the response, back into a response for each client.

PR #481, opened 2025-10-30
478 kafka 41 message definitions

### LINKED ISSUE #478: kafka 4.1 message definitions
- [x] in kafka 4.1 a couple of apis are marked with `version = "none"` that are no longer used that the proc macro isn't dealing with

shortishly:


solace-aross (changes requested):
Quick note before the review: I'm still fairly new to contributing here and learning the codebase, so please treat this as a helpful second pair of eyes rather than a final word — happy to be corrected on anything. I wasn't requested as a reviewer on this one, but it's been open a while with no review yet, so I dug in.

The message-descriptor work itself looks right. I checked the `DescribeConfigsResponse.json` change against the real Kafka 4.0/4.1 protocol change: v0 was dropped, `validVersions` correctly moves from `0-4` to `1-4`, and `IsDefault` (a v0-only field) is correctly removed. `cargo build -p tansu-sans-io` succeeds, so the generated code is internally consistent, and the diff stays within the intended boundary — only `tansu-model/src/lib.rs` and `tansu-sans-io/build.rs` are hand-edited, everything else is `message/*.json`.

The problem is that a few of the 4.1 shape changes aren't propagated to the code that consumes them, so the build is currently broken for every storage backend. I confirmed by running `cargo check -p tansu-storage --features postgres,dynostore,libsql` in a worktree at this PR's head:

- `tansu-storage/src/pg.rs:2481`, `dynostore.rs:1599`, `lite.rs:3147` — all call `.is_default(None)` on `DescribeConfigsResourceResult`, which no longer exists now that `IsDefault` is gone from the JSON. (`limbo.rs:2516` has the same call but is gated behind the `turso` feature, which isn't in the CI matrix — so it's a real but currently-latent break.)
- `tansu-storage/src/service/list_offsets.rs:161` — `old_style_offsets` no longer exists on `ListOffsetsPartitionResponse`.
- `tansu-storage/src/service/produce.rs:227` — type mismatch, looks like `topic.name` is now `Option<String>` where a `String` is expected.
- `tansu-storage/src/lib.rs:617` — `commit_timestamp` no longer exists on `OffsetCommitRequestPartition`.

That lines up exactly with this PR's own CI: all three `build-storage` legs (dynostore, libsql, postgres) and the `build-storage-lake` legs are red, while checks that don't touch storage (`src`, `typos`, `CodeQL`) are green.

Once those call sites are updated, the test suites will need a pass too — `tansu-sans-io/tests/{encode,decode,doctest,describe_configs}.rs`, `tansu-broker/tests/describe_configs.rs`, `tansu-schema/tests/berg.rs`, and the sixteen `.is_default(...)` calls in `tansu-schema/src/lake/delta.rs` (test-only, behind `#[cfg(test)]`) all still reference the removed field.

One more thing worth flagging, separate from the above: this PR is based on the pre-rename crate layout (`tansu-model`, `tansu-sans-io`, etc.), and `main` has since moved to the renamed `nisshi-*` layout. So beyond fixing the compile errors, it'll need a rebase to pick up the new paths — no fault of the PR, just worth knowing going in given how much has moved since this was opened.

Happy to take a look again once it's rebased and the storage-side call sites are updated — the underlying protocol work looks solid, it's really just the last mile of propagating the shape changes through.

PR #491, opened 2025-10-31
Bump the rust to 1.91

randyli:


randyli:
It seems that there's a permission problem. @shortishly

shortishly:
> It seems that there's a permission problem. @shortishly

firstly, thanks for making the change! 🍾 

yes, I don't want the "publish" part of the "package" step to run in this case.... need to work out how.

I'll merge your change shortly, thanks again 😄

PR #498, opened 2025-11-12
readme: add instructions on how to install locally and fix stale telemetry url docs

stanislavkozlovski:


README.md
```diff
@@ -347,11 +374,22 @@ Consumer:
   -b localhost:9092
 ```
 
-## Feedback
+## Local
+
+```shell
+cargo install --path tansu --all-features
```
solace-aross, line 380: This repo's crate directory is `nisshi` (see `nisshi/Cargo.toml`, `name = "nisshi"`), not `tansu` — there's no `tansu` subdirectory in the workspace, so `cargo install --path tansu --all-features` will fail with "no such file or directory" right after the clone. Should be `cargo install --path nisshi --all-features`.

README.md
```diff
@@ -347,11 +374,22 @@ Consumer:
...
+~/.cargo/bin/tansu -h
```
solace-aross, line 381: Same naming issue: the installed binary is named `nisshi`, so this should be `~/.cargo/bin/nisshi -h`.

solace-aross (commented):
Thanks for this, the quickstart and install sections are a nice addition and the OTLP flag fix is spot on.

I'm new to contributing here and still learning the codebase, so please take this in that spirit, but I did want to flag one thing I checked concretely rather than just read past: the Local install steps use the old `tansu` naming, and that will fail as written.

PR #501, opened 2025-11-14
wip: create a topic using slatedb storage

shortishly:
api:
- [x] create topic
- [ ] produce
- [ ] ...

using:

```shell
STORAGE_ENGINE="slatedb://tansu/"
```

create a topic using:

```shell
just customer-topic-create
```

levelled SST created in S3, looks like it could be a good fit for record batches and metadata:

```shell
pmorgan@Mac-mini tansu % just minio-mc ls -r local/tansu/tansu-tansu.slatedb    
docker compose exec minio /usr/bin/mc ls -r local/tansu/tansu-tansu.slatedb
[2025-11-14 15:10:24 UTC]   106B STANDARD manifest/00000000000000000001.manifest
[2025-11-14 15:10:24 UTC]   114B STANDARD manifest/00000000000000000002.manifest
[2025-11-14 15:10:24 UTC]   122B STANDARD manifest/00000000000000000003.manifest
[2025-11-14 15:20:19 UTC]   122B STANDARD manifest/00000000000000000004.manifest
[2025-11-14 15:20:19 UTC]   122B STANDARD manifest/00000000000000000005.manifest
[2025-11-14 15:22:20 UTC]   122B STANDARD manifest/00000000000000000006.manifest
[2025-11-14 15:22:20 UTC]   122B STANDARD manifest/00000000000000000007.manifest
[2025-11-14 15:10:24 UTC]    74B STANDARD wal/00000000000000000001.sst
[2025-11-14 15:20:19 UTC]    74B STANDARD wal/00000000000000000002.sst
[2025-11-14 15:20:30 UTC]   607B STANDARD wal/00000000000000000003.sst
[2025-11-14 15:22:20 UTC]    74B STANDARD wal/00000000000000000004.sst
```

PR #515, opened 2025-12-08
500 slatedb storage engine

jiaoew1991:
I resubmitted a PR based on https://github.com/tansu-io/tansu/pull/501

### DIFF
--- a/tansu-storage/src/lib.rs
+++ b/tansu-storage/src/lib.rs
@@ -2911,6 +2911,9 @@ impl Storage for StorageContainer {
 
             #[cfg(feature = "turso")]
             Self::Turso(engine) => engine.ping(),
+
+            #[cfg(feature = "slatedb")]
+            Self::Slate(engine) => engine.ping(),
         }
         .await
         .inspect(|_| {
--- a/tansu-storage/src/slate/storage.rs
+++ b/tansu-storage/src/slate/storage.rs
@@ -2375,4 +2375,10 @@ impl Storage for Engine {
     async fn advertised_listener(&self) -> Result<url::Url> {
         Ok(self.advertised_listener.clone())
     }
+
+    async fn ping(&self) -> Result<()> {
+        // Verify connectivity by attempting a simple read operation
+        let _ = self.db.get(Self::BROKERS).await?;
+        Ok(())
+    }
 }


jiaoew1991:
> We'll need to update the broker unit tests so that they include the slatedb storage engine:
> 
> There are a bunch of tests under "tansu-broker/tests" that are replicated for each storage engine. For example:
> 
> https://github.com/jiaoew1991/tansu/blob/4aaba29db9f7c9bd988d99470c2d6b5329396dd8/tansu-broker/tests/topic.rs#L473
> 
> Are the "topic" tests for SQLite, duplicating `lite` module for slate and then verify that it works the same as the other engines.

@shortishly Thank you for your suggestion, I have submitted a new commit to fix the comments 😄

jiaoew1991:
Hi @shortishly , All CI except smoke have already run, do you any more comments 😄

jiaoew1991:
Hi @shortishly Sorry for the late reply. Some basic issues have been fixed, but there may still be some points to consider in the use of slatedb. Slatedb only supports single-writer writes, which I think is fine in a single-broker scenario, but there is no good solution yet for a multi-broker scenario.

> comments around "Global Metadata Contention" and "optimistic concurrency control": slatedb is a single writer, have you observed any of theses issues with slatedb? I would expect (and do see) issues with the "plain" S3 engine because it uses optimistic locking, but not for slatedb? Also, have you thought about not having a single "topic metadata" but instead having one per topic? Retrieving the list of topics would be a scan of all matching keys rather than loading a single topic tree and returning that. Similarly for some of the other metadata (tx...)

I also agree with this point, it hasn't been changed yet. I think scanning according to the prefix would be better, and the performance of scanning under the same prefix is also quite good.

> * slatedb also supports partitioning of writers, which might be another reason to split out the global metadata so that it is "per topic". main issue is that a tx can enlist multiple topics, which could be in different partitions

You mentioned the partitioning of writers; I couldn't find any related information in the slatedb

jiaoew1991:
> @jiaoew1991 I'm looking to make a 0.6.0 release soon(ish) potentially with slatedb as a alpha/beta storage engine (and so that you're not merging out of main with this one all the time!). how is it looking? anything outstanding or that you're concerned about?

that's great🔥, This is currently the alpha version. I will continue to perform benchmarking and internal testing to verify it

shortishly:
@jiaoew1991 can you give this comment a 👍 if you're happy for me to merge this PR now?

We can then raise separate issues/PR with any followup (if necessary)

jiaoew1991:
> @jiaoew1991 can you give this comment a 👍 if you're happy for me to merge this PR now?
> 
> We can then raise separate issues/PR with any followup (if necessary)

👍🏻

PR #519, opened 2025-12-12
turso storage engine

### LINKED ISSUE #431: turso v0.2.0

shortishly:


solace-aross (changes requested):
This one should close rather than merge, though there's good material to carry forward in a smaller follow-up.

I'm still fairly new to this codebase, so before landing on that I checked the history carefully and had a second, more adversarial pass verify it. Details below.

**Why closing is the right call:**
- Main already ships the turso backend as a default feature (`nisshi-storage/Cargo.toml:14,19`) and has moved past this PR's target version: `turso = "0.7.1"` on main vs `"0.5.0"` here.
- `limbo.rs` has grown from 3999 lines in this PR to 4346 on main, with real fixes landed since the merge base (`279bc4a` "upgrade turso as result", `dc7c49a`).
- A test merge in a scratch worktree hits real conflicts in `Cargo.toml`, `Cargo.lock`, and `limbo.rs`. GitHub's own mergeable check agrees: `CONFLICTING`.

**What's still worth keeping, in a smaller follow-up PR:**
- `not null` on the autoincrement primary keys across the DDL files (e.g. `010-cluster.sql`). Main still doesn't have this.
- Dropping the trailing semicolon in `sql/topition_insert.sql`'s `returning` clause, which turso's parser needs.
- The `broker-turso` justfile recipe.
- Un-ignoring the turso tests in `nisshi-broker/tests/topic.rs`, though `limbo.rs:4250` on main notes turso/limbo still can't run the full produce flow in CI, and there's no turso CI job yet, so that piece may need to wait.

One thing worth leaving out of the follow-up: this PR also drops `not null` from the FK `cluster` columns (`020-topic.sql:17` and similar), which loosens a constraint shared with the libSQL backend. Worth checking whether that was intentional.

Thanks for pushing turso forward through several rebases. Glad to look at the follow-up whenever it's ready.

PR #542, opened 2026-01-04
refactor: move vacuum maintenance SQL to external file to fix cache misuse

RoomaPerera:
This PR addresses the err=LibSql(Misuse("Unknown cache key: select freelist_count, page_size FROM pragma_freelist_count(), pragma_page_size()")) warning in the log

Previously, the maintenance logic was passing a raw SQL string to a query function that expects a registered cache key. I have moved the SQL statement into its own file and registered it in the storage mapping to follow the project's established patterns.

#### Changes
- Created tansu-storage/src/sql/pg/maintain-vacuum.sql containing the freelist and page size query.
- Registered the new SQL file in the SQL cache mapping within tansu-storage/src/sql/mod.rs.
- Updated the query call in the maintenance logic to use the "maintain-vacuum.sql" key.

#### Verification:
- Verified that the filename, macro path, and query call all match exactly (maintain-vacuum.sql).
- Confirmed that the project compiles locally with cargo build -p tansu-storage

### DIFF
--- a/tansu-storage/src/sql.rs
+++ b/tansu-storage/src/sql.rs
@@ -63,8 +63,8 @@ macro_rules! include_sql {
 pub(crate) static SQL: LazyLock<Cache> = LazyLock::new(|| {
     let mapping = [
         (
-            "maintain_vacuum.sql",
-            include_sql!("pg/maintain_vacuum.sql"),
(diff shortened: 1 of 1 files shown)


shortishly:
thanks for picking this up @RoomaPerera 👍

PR #543, opened 2026-01-04
pretty-up README header

stanislavkozlovski:
This patch pretties up the first section a user sees in the readme, by adding a title in the center of it, an emoji, and a few useful links/badges


stanislavkozlovski:
# Before

image


# After

image

PR #545, opened 2026-01-06
Topic command error handling #346

shaneikennedy:
Started to pick up #346, the first error I tried to handle was the case when the broker is unavailable and I realized that the ConnectionManager is using the default "timeout" policy from the backoff crate which is particularly long (15m). The first commit in this PR is to set a much smaller value for this behavior (30s). Let me know if this needs reworking, I think a small timeout makes sense for an admin cli command like `topic create` but might not make sense for all the client interactions. If it's preferable I could make this configurable on the ConnectionManager, i actually had a previous attempt that included this but i removed it in case it was overkill and unwanted 👍 

The second commit is to actually handle the various errors that can happen in the tansu topic commands:
- For anything that maps to a ErrorCode from the tansu-sans-io package, but isn't the None variant, just error log because the ErrorCode implements display
- For specifically a connection problem I handle that as well

Another more all-encompassing way to make cli error reporting better would be to handle the error code logic in tansu/src/bin/tansu.rs
```
    Cli::main()
        .await
        .inspect(|error_code| match error_code {
                ErrorCode::None => info!("Topic command exited successfully."),
                _ => error!("{}", error_code))
        .inspect_err(|err| error!(%err))
```

and then in the topic command logic I can handle the connection error diagnostic specifically

lmk what you think here, happy to work on this with a bit more direction 👍

### DIFF
--- a/tansu-cli/src/cli/topic.rs
+++ b/tansu-cli/src/cli/topic.rs
@@ -18,6 +18,7 @@ use crate::Result;
 use clap::Subcommand;
 use tansu_sans_io::ErrorCode;
 use tansu_topic::Topic;
+use tracing::{debug, error, info};
 use url::Url;
 
 use super::DEFAULT_BROKER;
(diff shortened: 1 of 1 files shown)


shortishly:
Very cool. Thanks very much for this.

PR #568, opened 2026-01-20
feat: support lance as lake storage

jiaoew1991:


tansu-schema/src/lake/lance.rs
```diff
@@ -0,0 +1,557 @@
...
+            .await?;
+
+        let num_rows = record_batch.num_rows();
+        debug!(%topic, partition, offset, rows = num_rows, columns = record_batch.num_columns());
+
+        // Convert to lance's arrow version (v56)
+        let lance_batch = convert_record_batch(&record_batch)?;
+
+        // Try to open or create the dataset
+        let _ = self.open_or_create_dataset(topic, &lance_batch).await?;
```
solace-aross, line 267: I think this double-writes the first batch for any brand-new topic.

`open_or_create_dataset` (around line 152-169 above) doesn't just check for existence: when the dataset doesn't exist yet, it writes the given batch as part of creating it, via `Dataset::write(reader, &uri, WriteParams { mode: WriteMode::Create, .. })`. That write already persists the rows. The returned dataset is discarded here (`let _ = ...`), and the very next line calls `append()`, which builds a fresh reader from the *same* `lance_batch` and writes it again via `WriteMode::Append`.

Net effect: the first `store()` call for any new topic writes that batch's rows twice, permanently. Checked this against the vendored lance 1.0.1 source — `Dataset::write` calls `write_fragments_internal` unconditionally before it even inspects `mode`; `Create` vs `Append` only changes the resulting commit operation (`Overwrite` vs `Append`), not whether the rows get written. So both calls here persist data.

Worth noting: every row-count assertion in `tests/lance.rs`, and in this file's own `#[cfg(test)]` module, checks `num_rows() > 0` rather than an exact count, so the existing suite passes either way. Might be worth tightening a couple of those to an exact count once this is fixed, so a regression here doesn't go unnoticed again.

Smallest fix I can see: have `open_or_create_dataset` signal whether it just created the dataset (it already wrote the batch in that case) versus opened an existing one, and skip the trailing `append()` on the create path.

tansu-schema/src/lake/lance.rs
```diff
@@ -0,0 +1,557 @@
...
+            return Ok(dataset);
+        }
+
+        // Try to open existing dataset
+        let dataset = match Dataset::open(&uri).await {
+            Ok(dataset) => {
+                debug!(opened = ?uri);
+                dataset
+            }
+            Err(_) => {
```
solace-aross, line 157: Non-blocking observation: this treats *any* error from `Dataset::open` (not just "not found") as "dataset doesn't exist yet" and falls through to create it. A transient I/O or permission error would take the same path and could surface as a confusing `DatasetAlreadyExists` (or worse) instead of the real underlying error. Might be worth narrowing this to the specific "not found" case if the lance error type supports matching on it, so real failures don't get masked.

solace-aross (changes requested):
I'm still new to this codebase, so treat this as a good-faith second look rather than gospel, and please push back if I've got something wrong.

Lance support fits the existing `House`/`LakeHouse` pattern well: feature flag, builder, and trait impl all line up cleanly with delta/iceberg. Two things I think need a look before this merges:

1. **CI is currently red.** `build-storage-lake` fails for the `lance` feature on all four storage backends (dynostore, libsql, postgres, slatedb) — worth checking the checks tab.
2. **Likely data duplication on the first write to any new topic.** Details in the inline comment on `lance.rs`.

Also left one non-blocking observation below. Thanks for putting this together, happy to help however's useful as I get more familiar with the codebase.

PR #637, opened 2026-03-26
default retry configuration

### LINKED ISSUE #636: dynostore retry configuration
https://docs.rs/object_store/latest/object_store/struct.RetryConfig.html

shortishly:


tansu-storage/src/lib.rs
```diff
@@ -1742,6 +1745,7 @@ impl Builder<i32, String, Url, Url> {
                 AmazonS3Builder::from_env()
                     .with_bucket_name(bucket_name)
                     .with_conditional_put(S3ConditionalPut::ETagMatch)
+                    .with_retry(RetryConfig::default())
```
solace-aross, line 1748: Small note: `object_store`'s `AmazonS3Builder` derives `Default`, and its `retry_config` field is already `RetryConfig::default()` unless `.with_retry(...)` is called, same values this sets explicitly (`max_retries: 10`, `retry_timeout: 3 min`). There's also no `AWS_*` config key that touches retry, so this line doesn't change runtime behavior today.

Could still be worth keeping, e.g. as an explicit anchor for later making retry configurable via the storage URL. The `batch_min_size`/`batch_max_delay` query-param pattern used a few lines above for the S3 and GCS builders looks like a natural fit. Just flagging so the intent is clear either way.

solace-aross (approved):
Approving, thanks for a small, clean PR to dig into.

Verified it merges and builds cleanly against current `main` in a scratch worktree, and CI is green (52/52). Left one observation inline on the `with_retry` call, not a blocker, just something I noticed while checking `object_store`'s internals.

I'm still learning this codebase, so please read that note as a question more than a correction. Happy to be told I've got it wrong.

PR #661, opened 2026-04-28
fix(storage): strip compression bits from rebuilt batches in fetch path

erikcw:
## Summary

The libsql/`lite.rs`, turso/`limbo.rs`, and postgres/`pg.rs` storage backends decompose record batches into individual rows on produce (after inflating the batch) and reconstruct fresh batches from those rows on fetch. The reconstruction copied the original batch attributes verbatim, including the compression bits, even though the stored records are already decompressed. The subsequent ` inflated -> deflated` conversion at the end of fetch then failed for any compression type not handled on the deflate path -- in particular Snappy, which `tansu_sans_io::record::deflated::into_record_data` does not implement, surfacing as `SansIo(UnexpectedType("Snappy"))` and silently dropping batches from consumers (e.g. franz-go producers with `kgo.SnappyCompression()`).

We hit this in production with the SQLite backend, where the broker log accumulated hundreds of `UnexpectedType("Snappy")` lines per session and the consumer never received the messages.

## Fix

Mask off the compression bits (`& !0b111`) when reading attributes into the rebuilt batch so the inflated -> deflated conversion always sees `Compression::None`, matching the actual record-data state. Apply the same mask to the in-loop `attributes` value used for batch boundary detection so we don't fragment batches by compression bits that no longer apply.

`dynostore` (S3/memory) and `slatedb` store batches as opaque blobs and are unaffected; this PR does not touch them.

## Test plan

- New regression test `tansu-storage/tests/snappy_fetch_round_trip.rs`:
  - Hand-rolls a Snappy-compressed `deflated::Batch` (since the inflated -> deflated path doesn't yet support Snappy).
  - Produces it via the Storage trait, fetches it back, asserts the records decode to the original payload.
  - libsql variant runs by default; turso and postgres variants are wired up behind `#[ignore]` to match how the existing turso/postgres broker tests are gated.
- Existing `tansu-broker` `lite::*` produce_fetch tests still pass after the change.

Validated end-to-end against a live deployment (libsql backend, SCRAM-SHA-256, 5 Snappy-compressed records via `rpk topic produce --compression snappy` and `rpk topic consume`): unpatched broker times out on consume, patched broker returns all 5 records.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8562,6 +8562,7 @@ dependencies = [
  "serde_json",
  "sha2",
  "slatedb",
+ "snap",
  "tansu-sans-io",
  "tansu-schema",
  "tempfile",
--- a/tansu-storage/Cargo.toml
+++ b/tansu-storage/Cargo.toml
@@ -71,6 +71,7 @@ assert_matches.workspace = true
 criterion.workspace = true
 dotenv.workspace = true
 pretty_assertions.workspace = true
+snap.workspace = true
 tansu-schema.workspace = true
 tempfile.workspace = true
 time.workspace = true
--- a/tansu-storage/src/limbo.rs
+++ b/tansu-storage/src/limbo.rs
@@ -1511,7 +1511,15 @@ impl Storage for Engine {
                         .inspect_err(|err| error!(?err))?,
                 )
                 .attributes(
-                    row.get_value(1)
+                    // Records are stored decompressed (the produce
+                    // path inflates before insert), so strip the
+                    // compression bits when rebuilding the batch.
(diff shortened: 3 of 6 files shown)


shortishly:
@erikcw please take a look at #667. After inflating a batch the compression attribute is set to None. I think it should solve your issue without requiring changes on SQL based storage engines.

Please reopen if I've missed something.

shortishly:
> for any compression type not handled on the deflate path -- in particular Snappy, which `tansu_sans_io::record::deflated::into_record_data` does not implement, surfacing as

I seem to remember there was an issue with the Snappy encoder, this could do with being a separate issue.

PR #665, opened 2026-04-28
fix(auth): support SASL re-authentication on existing connections (KIP-368)

erikcw:
## Summary

`SaslHandshakeService` only proceeded when `Authentication`'s `Stage` was `Server`. After a successful initial handshake the Stage transitions \`Server -> Session -> Finished\` and never resets, so any subsequent \`SaslHandshakeRequest\` on the same TCP connection fell into the else branch and was rejected with `UnsupportedSaslMechanism` — a misleading error since the mechanism was supported, the state machine just refused to start a second exchange.

Java `kafka-clients` periodically issues `SaslHandshake` / `SaslAuthenticate` on existing connections per [KIP-368](https://cwiki.apache.org/confluence/display/KAFKA/KIP-368%3A+Allow+SASL+Connections+to+Periodically+Re-Authenticate). Strimzi Kafka Connect, the Confluent Avro converter, and any other long-lived Java producer/consumer therefore failed re-authentication against tansu with:

```
Failed re-authentication with broker (channelId=N) (Client SASL
mechanism 'SCRAM-SHA-256' not enabled in the server, enabled
mechanisms are [SCRAM-SHA-256])
```

The franz-go-based clients in `redpanda-connect` ran into the same wall on their re-auth path. `rpk` happens to work because its short-lived CLI invocations don't reach the re-auth interval.

We hit this trying to bring up Strimzi Kafka Connect against a tansu broker for an S3 sink workload — the connect worker reauths every minute and dies on the first attempt.

## Fix

Keep the `Arc<SASLConfig>` alongside `Authentication` and reissue a fresh `SASLServer` at the start of every handshake, discarding any prior `Stage`. The first-handshake path is unchanged; the second-and-subsequent handshakes now succeed.

\`AuthenticationFrame\` in \`tansu-service\` had a \`#[derive(Default)]\` it never relied on; that derive is dropped to accommodate the new required \`config\` field on \`Authentication\`.

## Test plan

- New regression test \`auth_handshake_scram_256_v1_reauth\` in \`tansu-broker/tests/auth.rs\`:
  - Runs the full SCRAM-SHA-256 handshake + multi-step authenticate exchange.
  - Then issues a second \`SaslHandshakeRequest\` on the same broker service (same \`Authentication\`).
  - Asserts the second handshake's \`error_code\` is \`None\`.
  - Pre-fix this fails at the assertion with \`right: UnsupportedSaslMechanism\`.
  - Post-fix it passes.
- Existing \`auth_handshake_scram_256_v1\`, \`auth_handshake_scram_512_v1\`, \`auth_handshake_scram_512_bad_password_v1\`, \`not_authenticated\`, and \`auth_handshake_scram_256_v0\` tests continue to pass.

Validated end-to-end against a live broker (libsql backend, SCRAM-SHA-256, Strimzi Kafka Connect 0.51 with kafka-clients 4.x): pre-fix every long-lived Java consumer dies in CrashLoopBackOff with the re-auth error after the first re-auth interval; post-fix consumers stay healthy across multiple re-auth cycles.

### DIFF
--- a/tansu-auth/src/handshake.rs
+++ b/tansu-auth/src/handshake.rs
@@ -43,9 +43,20 @@ where
             .lock()
             .map_err(Into::into)
             .and_then(|mut guard| {
-                if let Some(Stage::Server(server)) = guard.take()
-                    && let Ok(mechanism) = Mechname::parse(req.mechanism.as_bytes())
-                {
+                // Re-authentication (KIP-368): a Java kafka-clients
+                // connection periodically issues another SaslHandshake
+                // on the same TCP socket. The previous handshake left
+                // the Stage in `Session`/`Finished`, so a stale
+                // `take()` would fall into the else branch and reject
+                // a perfectly valid mechanism with
+                // `UnsupportedSaslMechanism`. Always start with a
+                // fresh `SASLServer` built from the stored config.
+                _ = guard.take();
+                let Stage::Server(server) = authentication.fresh_server() else {
+                    unreachable!("fresh_server returns Stage::Server")
+                };
+
+                if let Ok(mechanism) = Mechname::parse(req.mechanism.as_bytes()) {
                     debug!(available = ?server.get_available().into_iter().map(|mechanism|mechanism.mechanism.as_str()).collect::<Vec<_>>());
 
                     server
--- a/tansu-auth/src/handshake.rs
+++ b/tansu-auth/src/handshake.rs
@@ -51,12 +51,12 @@ where
                 // a perfectly valid mechanism with
                 // `UnsupportedSaslMechanism`. Always start with a
                 // fresh `SASLServer` built from the stored config.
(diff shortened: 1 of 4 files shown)


shortishly:
@erikcw thanks very much for this. I've made a (hopefully) minor change to remove unreachable. I'll merge this once CI has completed.

shortishly:
> We hit this trying to bring up Strimzi Kafka Connect against a tansu broker for an S3 sink workload — the connect worker reauths every minute and dies on the first attempt.

The default used by tansu:

https://github.com/tansu-io/tansu/blob/fadbe28ce7ef71ad760a0aab8e8e8ae57ee1ae41/tansu-auth/src/authenticate.rs#L33

I haven't been able to find the default used by apache kafka for this. 1m is too short.

PR #689, opened 2026-06-24
fix(storage): stop read_committed from reading past the last stable offset

OrHayat:
The watermark query selects three values:
    select w.low, w.high, s.offset

For the read_committed boundary (last_stable), offset_stage picked w.high -- the end of the log -- instead of s.offset, the stable offset up to which everything is committed. So read_committed consumers were handed the end of the log and could read records from transactions that hadn't committed yet.

Read s.offset instead on all three backends (Postgres, libSQL, Turso), with a regression test on each (turso one is  disabled like other tests until turso is stablized)

### DIFF
--- a/tansu-storage/src/limbo.rs
+++ b/tansu-storage/src/limbo.rs
@@ -1743,7 +1743,7 @@ impl Storage for Engine {
             .unwrap_or_default();
 
         let last_stable = row
-            .get_value(1)
+            .get_value(2)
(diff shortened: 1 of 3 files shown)


shortishly:
@OrHayat thanks for this 👍 

I'm pulling together the next release, with some more integration tests, and this will help with some of the TX issues found by those tests... the main branch will be a little unstable while I pull various changes together.

I'll merge your change once the CI process has completed.

PR #690, opened 2026-06-24
Add caching to Dockerfile

simon-engledew:
I found myself rebuilding this image quite a lot while I was working on fixing an s3 performance issue. 😅 

This tweaks the Dockerfile to be more rebuild friendly.

I guess there is a good chance you wont want or need this change, in which case no worries! If nothing else it might be helpful for anyone else who needs to `docker build` this locally a bunch of times in a row. 😁

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1
 # Copyright ⓒ 2024-2026 Peter Morgan <peter.james.morgan@gmail.com>
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -17,25 +18,35 @@ FROM --platform=$BUILDPLATFORM tonistiigi/xx AS xx
 FROM --platform=$BUILDPLATFORM rust:1.88-alpine AS builder
 COPY --from=xx / /
 RUN apk add clang cmake lld
-RUN rustup target add $(xx-cargo --print-target-triple)
-
-WORKDIR /usr/src
-ADD / /usr/src/
 
 ARG TARGETPLATFORM
 RUN xx-apk add --no-cache musl-dev zlib-dev zlib-static gcc
-RUN xx-cargo build --bin tansu --all-features --release --target-dir ./build
-RUN xx-verify --static ./build/$(xx-cargo --print-target-triple)/release/tansu
 
-RUN <<EOF
-mkdir -p /image/schema /image/data /image/tmp /image/etc/ssl
-cp -v build/$(xx-cargo --print-target-triple)/release/tansu /image
-cp -v LICENSE /image
-cp -rv /etc/ssl /image/etc
+WORKDIR /usr/src
+
+COPY rust-toolchain.toml .
+RUN rustup show && rustup target add $(xx-cargo --print-target-triple)
+
+ADD / /usr/src/
+
+ARG CARGO_BUILD_JOBS=default
+ENV CARGO_BUILD_JOBS=$CARGO_BUILD_JOBS
+
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    --mount=type=cache,target=/usr/local/cargo/git \
+    --mount=type=cache,target=/usr/src/build <<EOF
+xx-cargo build --bin tansu --no-default-features --features dynostore --release --target-dir ./build
+xx-verify --static ./build/$(xx-cargo --print-target-triple)/release/tansu
+install ./build/$(xx-cargo --print-target-triple)/release/tansu /usr/bin
 EOF
 
+FROM scratch AS out
+
+COPY --from=builder --parents /etc/ssl /
+COPY --from=builder /usr/src/LICENSE /usr/bin/tansu /
+
 FROM scratch
-COPY --from=builder /image /
+COPY --from=out / /
 ENV TMP=/tmp
 ENTRYPOINT ["/tansu"]
 CMD ["broker"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,9 +32,9 @@ ADD / /usr/src/
 ARG CARGO_BUILD_JOBS=default
 ENV CARGO_BUILD_JOBS=$CARGO_BUILD_JOBS
 
-RUN --mount=type=cache,target=/usr/local/cargo/registry \
-    --mount=type=cache,target=/usr/local/cargo/git \
-    --mount=type=cache,target=/usr/src/build <<EOF
+RUN --mount=type=cache,target=/usr/local/cargo/registry,id=cargo-registry-${TARGETPLATFORM},sharing=locked \
+    --mount=type=cache,target=/usr/local/cargo/git,id=cargo-git-${TARGETPLATFORM},sharing=locked \
+    --mount=type=cache,target=/usr/src/build,id=cargo-build-${TARGETPLATFORM} <<EOF
 xx-cargo build --bin tansu --no-default-features --features dynostore --release --target-dir ./build
 xx-verify --static ./build/$(xx-cargo --print-target-triple)/release/tansu
 install ./build/$(xx-cargo --print-target-triple)/release/tansu /usr/bin


Dockerfile
```diff
@@ -17,25 +18,35 @@ FROM --platform=$BUILDPLATFORM tonistiigi/xx AS xx
...
-RUN xx-cargo build --bin tansu --all-features --release --target-dir ./build
-RUN xx-verify --static ./build/$(xx-cargo --print-target-triple)/release/tansu
-
-RUN <<EOF
-mkdir -p /image/schema /image/data /image/tmp /image/etc/ssl
-cp -v build/$(xx-cargo --print-target-triple)/release/tansu /image
-cp -v LICENSE /image
-cp -rv /etc/ssl /image/etc
+ARG CARGO_BUILD_JOBS=default
+ENV CARGO_BUILD_JOBS=$CARGO_BUILD_JOBS
```
simon-engledew, line 33: Without setting this to 4 the LTO causes cargo to OOM on my laptop, even after I give it a lot of RAM. 😞

Dockerfile
```diff
@@ -17,25 +18,35 @@ FROM --platform=$BUILDPLATFORM tonistiigi/xx AS xx
...
+    --mount=type=cache,target=/usr/src/build <<EOF
+xx-cargo build --bin tansu --no-default-features --features dynostore --release --target-dir ./build
+xx-verify --static ./build/$(xx-cargo --print-target-triple)/release/tansu
+install ./build/$(xx-cargo --print-target-triple)/release/tansu /usr/bin
 EOF
 
+FROM scratch AS out
+
+COPY --from=builder --parents /etc/ssl /
+COPY --from=builder /usr/src/LICENSE /usr/bin/tansu /
```
simon-engledew, line 46: I tweaked this so it still only generates one layer in the final image but I found it easier to see when a file was missing using COPY instead of cp. 🤷 

I wouldn't be surprised if docker didn't squash multiple copy instructions automatically these days, so this might be totally unnecessary. 😅

Dockerfile
```diff
@@ -17,25 +18,35 @@ FROM --platform=$BUILDPLATFORM tonistiigi/xx AS xx
...
 RUN apk add clang cmake lld
-RUN rustup target add $(xx-cargo --print-target-triple)
+
+ARG TARGETPLATFORM
+RUN xx-apk add --no-cache musl-dev zlib-dev zlib-static gcc
 
 WORKDIR /usr/src
+
+COPY rust-toolchain.toml .
+RUN rustup show && rustup target add $(xx-cargo --print-target-triple)
```
simon-engledew, line 28: I moved this before the source ADD so that it doesn't have to be redownloaded/installed on every source tweak.

Dockerfile
```diff
@@ -17,25 +18,35 @@ FROM --platform=$BUILDPLATFORM tonistiigi/xx AS xx
...
-cp -rv /etc/ssl /image/etc
+ARG CARGO_BUILD_JOBS=default
+ENV CARGO_BUILD_JOBS=$CARGO_BUILD_JOBS
+
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    --mount=type=cache,target=/usr/local/cargo/git \
+    --mount=type=cache,target=/usr/src/build <<EOF
+xx-cargo build --bin tansu --no-default-features --features dynostore --release --target-dir ./build
+xx-verify --static ./build/$(xx-cargo --print-target-triple)/release/tansu
+install ./build/$(xx-cargo --print-target-triple)/release/tansu /usr/bin
```
simon-engledew, line 40: You have to copy this out of the build cache or it wont be in the layer for the copy stage.

shortishly:
Nice, will take a look

Dockerfile
```diff
@@ -17,25 +18,35 @@ FROM --platform=$BUILDPLATFORM tonistiigi/xx AS xx
...
-mkdir -p /image/schema /image/data /image/tmp /image/etc/ssl
-cp -v build/$(xx-cargo --print-target-triple)/release/tansu /image
-cp -v LICENSE /image
-cp -rv /etc/ssl /image/etc
+ARG CARGO_BUILD_JOBS=default
+ENV CARGO_BUILD_JOBS=$CARGO_BUILD_JOBS
+
+RUN --mount=type=cache,target=/usr/local/cargo/registry,id=cargo-registry-${TARGETPLATFORM},sharing=locked \
+    --mount=type=cache,target=/usr/local/cargo/git,id=cargo-git-${TARGETPLATFORM},sharing=locked \
+    --mount=type=cache,target=/usr/src/build,id=cargo-build-${TARGETPLATFORM} <<EOF
```
simon-engledew, line 37: Ah. I've changed the cache mount to include TARGETPLATFORM so that when CI builds the two images at the same time they don't write over each other. 😅

Dockerfile
```diff
@@ -17,25 +18,35 @@ FROM --platform=$BUILDPLATFORM tonistiigi/xx AS xx
...
+FROM scratch AS out
+
+COPY --from=builder --parents /etc/ssl /
+COPY --from=builder /usr/src/LICENSE /usr/bin/tansu /
+
 FROM scratch
-COPY --from=builder /image /
+COPY --from=out / /
 ENV TMP=/tmp
 ENTRYPOINT ["/tansu"]
```
solace-aross: This still points at `/tansu`. Confirmed locally that this branch doesn't contain the `tansu` -> `nisshi` rename commit (7bbbb75) yet, so the binary name, verify path, install path, and COPY on lines 38-46 above are all stale the same way. Rebasing onto `main` should clear this up.

solace-aross (commented):
Thanks for this, a friendlier local rebuild loop is a real quality-of-life win, and the caching approach (BuildKit cache mounts, moving toolchain setup earlier, multi-stage `scratch` output) is a solid design.

One thing worth sorting out before this merges: the branch predates the `tansu` -> `nisshi` rename (commit 7bbbb75, 2026-06-27), and GitHub is now showing it as `CONFLICTING`/`DIRTY`. I checked out the branch locally to confirm: the Dockerfile still references `/tansu` in five places (build target, verify path, install path, COPY, ENTRYPOINT), and the source itself still builds the `tansu-sans-io` crate rather than `nisshi-sans-io`. None of the caching changes touch naming, so a rebase onto current `main` should be quick, just needed before this can land.

Small non-blocking question: the new final stage copies only `/etc/ssl` and the binary+license into `scratch`, dropping the old `mkdir -p /image/schema /image/data /image/tmp` step. `compose.yaml` bind-mounts `/schema` and `/data`, so Docker should create those mount points at container start regardless, and I didn't find any runtime code (only test helpers) relying on `$TMP`, so this is likely fine, just flagging in case it wasn't dropped on purpose.

The CI failures showing on this PR (postgres/sqlite compat, franz-go) look unrelated to the Dockerfile change, so I wouldn't hold this up on those.

I'm still fairly new to this codebase and learning as I go, so please take this as an extra pair of eyes rather than the final word. Happy to help however's useful!

Dockerfile
simon-engledew: Done!

simon-engledew:
> Small non-blocking question: the new final stage copies only /etc/ssl and the binary+license into scratch, dropping the old mkdir -p /image/schema /image/data /image/tmp step. compose.yaml bind-mounts /schema and /data, so Docker should create those mount points at container start regardless, and I didn't find any runtime code (only test helpers) relying on $TMP, so this is likely fine, just flagging in case it wasn't dropped on purpose.

Yeah, I dropped them because Docker should make them automagically and it seemed a little clunky to leak that implementation detail in the image. If you are planning on supporting people using the image in more esoteric ways perhaps creating these folders is necessary - in which case I could always add the mkdir back into the out step. 🤷

Dockerfile
```diff
@@ -17,25 +18,35 @@ FROM --platform=$BUILDPLATFORM tonistiigi/xx AS xx
...
-cp -v build/$(xx-cargo --print-target-triple)/release/nisshi /image
-cp -v LICENSE /image
-cp -rv /etc/ssl /image/etc
+ARG CARGO_BUILD_JOBS=default
+ENV CARGO_BUILD_JOBS=$CARGO_BUILD_JOBS
+
+RUN --mount=type=cache,target=/usr/local/cargo/registry,id=cargo-registry-${TARGETPLATFORM},sharing=locked \
+    --mount=type=cache,target=/usr/local/cargo/git,id=cargo-git-${TARGETPLATFORM},sharing=locked \
+    --mount=type=cache,target=/usr/src/build,id=cargo-build-${TARGETPLATFORM} <<EOF
+xx-cargo build --bin nisshi --no-default-features --features dynostore --release --target-dir ./build
```
solace-aross: This drops postgres/libsql/slatedb/turso (and the lake features) from the published image — see the review body for the trace. Restore the full feature set here, e.g. `--all-features`.

solace-aross (changes requested):
The caching approach itself is solid (BuildKit cache mounts for registry/git/build, moving toolchain setup earlier, multi-stage `scratch` output) and the rebase onto current `nisshi` naming is clean. One line changes what the published image can do, though, and needs to be fixed before this merges.

The build flag went from `--all-features` to `--no-default-features --features dynostore`. `dynostore` (`nisshi-storage/Cargo.toml:13-19`) only pulls in `object_store`; `postgres`, `libsql`, `slatedb`, and `turso` are separate, disjoint features, and `nisshi-storage/src/lib.rs` explicitly returns `Error::FeatureNotEnabled` for `postgres://`, `sqlite://`, `slatedb://`, and `turso://` when those features aren't compiled in (e.g. lines 2319-2323, 2463-2467, 2515-2519, 2534-2538). So the published `ghcr.io/nisshi-io/nisshi` image would only support `memory://` and `s3://` after this — no Postgres, no SQLite/libSQL, no SlateDB, no Turso, and none of `delta`/`iceberg`/`parquet` lake writing either, since those also dropped out of `--all-features`.

This Dockerfile is what `ci.yml`'s `package` job builds and pushes (`docker/build-push-action@v7`, `context: .`), and `smoke` pulls that same tag and runs it against `postgres://` and `sqlite://` (`ci.yml`'s smoke matrix). But neither of those would catch this on this PR: it's a fork PR awaiting maintainer approval to run any workflow at all, and even once approved, `smoke` is gated to same-repo PRs or pushes (`if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository`), so it never runs on a fork PR. This would land silently and only show up as a `smoke` failure against `main` after merge.

Can you restore the full feature set for the release build, e.g. back to `--all-features` (or the project's named `full` set: `delta,dynostore,iceberg,libsql,parquet,postgres,slatedb`)? None of the caching changes need the narrower feature set.

Dockerfile
simon-engledew: 🤦 Sorry, I must have committed this by accident when I was testing my binary search changes.

Dockerfile
```diff
@@ -17,25 +18,37 @@ FROM --platform=$BUILDPLATFORM tonistiigi/xx AS xx
...
+    --mount=type=cache,target=/usr/local/cargo/git,id=cargo-git-${TARGETPLATFORM},sharing=locked \
+    --mount=type=cache,target=/usr/src/build,id=cargo-build-${TARGETPLATFORM} <<EOF
+xx-cargo build --bin nisshi --all-features --release --target-dir ./build
+xx-verify --static ./build/$(xx-cargo --print-target-triple)/release/nisshi
+install ./build/$(xx-cargo --print-target-triple)/release/nisshi /usr/bin
 EOF
 
+FROM scratch AS out
+
+RUN mkdir -p /schema /data /tmp
```
solace-aross: Runs before any COPY, on a stage that's still bare `scratch`. No `/bin/sh` exists yet to exec `mkdir`. Reproduced the failure with podman/buildah (see review body). Create these directories in the `builder` stage and copy them across instead.

solace-aross (changes requested):
The feature-flag fix is right: `--all-features` is back, so the published image regains postgres/libsql/slatedb/turso and the lake features. Confirmed by diff against the prior review's flagged line.

One new line breaks the build, though: `RUN mkdir -p /schema /data /tmp` right after `FROM scratch AS out`, ahead of either `COPY`. `scratch` has no `/bin/sh`, so `RUN`'s default shell form has nothing to exec. Docker isn't available in my environment, so I reproduced the exact pattern with podman/buildah instead (same crun/runc OCI runtime path BuildKit shares):

```
error running container: from /usr/bin/crun creating container for
  [/bin/sh -c mkdir -p /schema /data /tmp]: executable file `/bin/sh` not found
```

This fails every build of this stage, including `package` in CI, which is what `smoke` then pulls and runs. Move the `mkdir` into the `builder` stage instead (same as `main` does today for `/image/schema` etc.), then let the existing `COPY --from=builder --parents` bring those directories across along with everything else.

Dockerfile
simon-engledew: Ok, that should have fixed it. Would it be possible to turn on CI so I can check the logs? Steam comes out of my laptop when I try to build the image locally...

solace-aross (approved):
The `mkdir`-on-`scratch` defect from the last round is fixed: switching to `WORKDIR` avoids needing a shell, so the build itself no longer fails. Confirmed the build succeeds (reproduced the earlier crun failure with the old `RUN mkdir` form, then confirmed the new `WORKDIR` form builds clean).

One thing this fix does *not* verify, and I want to flag rather than assume: whether `WORKDIR` on a `scratch` stage actually leaves `/schema`, `/data`, `/tmp` present in the resulting image, versus just succeeding at build time. I tested this directly with podman/buildah (`podman history --no-trunc` on an equivalent Dockerfile): every consecutive `WORKDIR` instruction produced a `0B` layer, and `podman save` + `tar -tv` on the final image showed no directory entries at all for `/schema` or `/data`, only the files that were explicitly `COPY`'d in.

I can't fully trust that result as a stand-in for real Docker/BuildKit, though — BuildKit added a dedicated no-exec `mkdir` op specifically so `WORKDIR` works on `scratch` (the same reason it doesn't need `/bin/sh` where `RUN mkdir` did), and buildah may not implement that same optimization, so my podman repro could be testing the wrong engine's behavior rather than a real defect in this PR.

Given that, I'm approving rather than blocking on it, because even in the worst case the impact is bounded: the build still succeeds either way, `compose.yaml`'s bind mounts for `/schema` and `/data` get auto-created by Docker at container start if missing, and (per my first-round check) no runtime code depends on `/tmp` pre-existing — only test helpers do. But it's cheap to settle for good: a 30-second `docker build --target out -o type=tar,dest=out.tar .` followed by `tar -tvf out.tar` would confirm the directories are really there under the engine CI actually uses. Worth doing before or shortly after merge, not worth reopening review for.

The caching changes themselves (BuildKit mounts for the cargo registry/git/build caches, moving toolchain setup earlier, the `--parents` copy) are solid and unchanged in substance since the first round.

solace-aross (approved):
Verified the new commit resolves the WORKDIR caveat from the earlier review.

I isolated just the directory-creation mechanism in two throwaway single-line Dockerfiles (not the full nisshi build, which doesn't need to change here) and compared `podman history` + `podman save` output:

- Old approach (`WORKDIR /schema` etc. on `FROM scratch`): every step shows `0B` / `#(nop)` in history, and there is no `layer.tar` at all for any of them in the saved image. WORKDIR only sets image config (`WorkingDir`) here — it never touched the filesystem, so `/schema`, `/data`, `/tmp` genuinely did not exist in the image that got approved earlier.
- New approach (`RUN --mount=from=busybox:musl,... mkdir /schema /data /tmp`): produces a real 4.1kB layer with actual `drwxr-xr-x` entries for `schema/`, `data/`, `tmp/` (plus `dev/`, `proc/`, `sys/` as a harmless side effect of the busybox mount — useful mountpoints for a container runtime to have anyway).

So this wasn't just a buildah-vs-BuildKit gap as flagged last pass — it's a genuine fix. Approving at this head.

Issue #684, opened 2026-06-08
S3 (dynostore): fetch lists the entire partition prefix on every call (O(total batches) per fetch)


ggekos:
## Summary

On the S3 / `dynostore` engine, `Storage::fetch` lists the **entire** partition's `records/` prefix and builds a set of **all** batch base-offsets on **every** call — regardless of the requested fetch offset. Each fetch therefore costs O(total number of batches in the partition) in S3 LIST calls, CPU and transient memory, instead of scaling with the amount of data actually returned.

## Where

`tansu-storage/src/dynostore.rs`, `fetch()` (≈ L1019–1086 on `main` / v0.6.0):

- it `object_store.list(records_prefix)` over the whole prefix and inserts every parsed base-offset into a `BTreeSet`,
- then `offsets.split_off(&offset)` selects the batches `>=` the requested offset and reads them until `max_bytes` is reached.

The data actually read is correctly bounded by `max_bytes`; the problem is the **full-prefix enumeration performed on each fetch**.

## Impact

- Fetch latency and S3 LIST cost grow with the partition's total history, not with the request size.
- A consumer that reads from (or seeks near) the start, and then issues successive fetches to advance, re-lists the whole — and growing — prefix on each fetch → roughly quadratic over a full read. On partitions with many `.batch` objects this becomes slow and memory-heavy.
- Transient memory per fetch scales with the number of batches in the partition (the base-offset set), independent of `max_bytes`.

## Reproduction (sketch)

Create a partition with a large number of `.batch` objects (many small/frequent produce requests). Reading the partition from the beginning to its end gets disproportionately slow and degrades further as the partition grows, even though each individual fetch only returns `max_bytes`.

## Proposed fix: `start-after`

Use S3 `start-after` to begin listing **near the requested offset** instead of at the start of the partition. Batch filenames are zero-padded to 20 digits, so lexicographic order == numeric order and `start-after` maps directly to numeric seeking:

- list with `object_store::ObjectStore::list_with_offset(records_prefix, start_after_key)` where `start_after_key` is derived from the requested offset, and stop once enough batches to satisfy `max_bytes` are collected;
- predecessor nuance: the requested offset may fall inside a batch whose `base_offset` is `< offset`, so the start key must be chosen to include that predecessor batch (today `split_off(&offset)` already assumes a batch boundary). In practice: `start-after` the key just below the predecessor base-offset.

[...]

PR #691, opened 2026-06-25
Minimal change to use binary search for dynostore fetch/list_offsets

simon-engledew:
This is an attempt to move fetch and list_offsets onto a binary search rather than a full bucket scan when using the s3 backend.

I've tried to keep the change as small as possible and add some test coverage. 🙇 I'm not terribly good at Rust so it's likely this is inaccurate.

It's hard for me to gauge if this is definitely correct wrt. the transaction semantics. I have at least verified this against a bucket with ~500mb of messages in and it is able to make progress after this change (and not before).

Full disclosure, I've had to make a bunch of further changes to get this running _well_ for said bucket (concurrently fetch and decode the s3 objects in parallel, a slightly more involved version of this binary search change) and I'm not running this minimal version of the change in anger anywhere.

x-ref: https://github.com/tansu-io/tansu/issues/684

Is this helpful/reviewable? 😅 

Please let me know if there is anything else I can do to make this easier to work with.

### DIFF
--- a/tansu-storage/src/dynostore.rs
+++ b/tansu-storage/src/dynostore.rs
@@ -1467,10 +1467,7 @@ impl Storage for DynoStore {
                             ListOffset::Latest => offset + 1,
                             _ => offset,
                         }),
-                        timestamp: match offset_request {
-                            ListOffset::Timestamp(_) => Some(found.last_modified.into()),
-                            _ => None,
-                        },
+                        timestamp: Some(found.last_modified.into()),
                     },
                 ))
             } else {
--- a/tansu-storage/src/dynostore.rs
+++ b/tansu-storage/src/dynostore.rs
@@ -497,6 +497,89 @@ impl DynoStore {
         }
     }
 
+    /// The metadata for the least batch at or after `probe` in the partition,
+    /// or `None` when no batch starts at or after it. Record objects are keyed
+    /// by zero-padded base offset, so a prefix listing that starts just before
+    /// `probe` yields the next base offset first.
+    async fn batch_meta_at_or_after(
+        &self,
+        topition: &Topition,
+        probe: i64,
+    ) -> Result<Option<ObjectMeta>> {
+        let location = Path::from(format!(
+            "clusters/{}/topics/{}/partitions/{:0>10}/records/",
+            self.cluster, topition.topic, topition.partition,
+        ));
+
+        let mut list_stream = if probe > 0 {
+            let start_after = Path::from(format!(
+                "clusters/{}/topics/{}/partitions/{:0>10}/records/{:0>20}.batch",
+                self.cluster,
+                topition.topic,
+                topition.partition,
+                probe - 1,
+            ));
+
+            self.object_store
+                .list_with_offset(Some(&location), &start_after)
+        } else {
+            self.object_store.list(Some(&location))
+        };
+
+        list_stream
+            .next()
+            .await
+            .transpose()
+            .inspect_err(|error| error!(?error, ?topition, probe))
+            .map_err(|_| Error::Api(ErrorCode::UnknownServerError))
+    }
+
+    /// The base offset encoded in a record object's key, or `None` when the key
+    /// has no terminal segment.
+    fn batch_base_offset(meta: &ObjectMeta) -> Result<Option<i64>> {
+        meta.location
+            .parts()
+            .next_back()
+            .map(|base| i64::from_str(&base.as_ref()[0..20]))
+            .transpose()
+            .map_err(Into::into)
+    }
+
+    /// The least batch base offset at or after `probe` in the partition, or
+    /// `None` when no batch starts at or after it.
+    async fn batch_base_at_or_after(&self, topition: &Topition, probe: i64) -> Result<Option<i64>> {
+        match self.batch_meta_at_or_after(topition, probe).await? {
+            Some(meta) => Self::batch_base_offset(&meta),
+            None => Ok(None),
+        }
+    }
+
+    /// The greatest batch base offset at or before `offset` in the partition.
+    /// Object listing is forward-only, so this is a binary search over
+    /// [`Self::batch_base_at_or_after`] successor probes.
+    async fn batch_base_at_or_before(
+        &self,
+        topition: &Topition,
+        offset: i64,
+    ) -> Result<Option<i64>> {
+        let mut floor = None;
+        let (mut lo, mut hi) = (0, offset);
+
+        while lo <= hi {
+            let mid = lo + (hi - lo) / 2;
+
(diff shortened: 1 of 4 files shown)


tansu-storage/src/dynostore.rs
```diff
@@ -497,6 +497,89 @@ impl DynoStore {
...
+            let mid = lo + (hi - lo) / 2;
+
+            match self.batch_base_at_or_after(topition, mid).await? {
+                Some(base) if base <= offset => {
+                    floor = Some(base);
+                    lo = base + 1;
+                }
+                _ => hi = mid - 1,
+            }
+        }
```
simon-engledew, line 583: This is the actual binary search.

tansu-storage/src/dynostore.rs
```diff
@@ -1051,7 +1134,28 @@ impl Storage for DynoStore {
...
+                let start_after = Path::from(format!(
+                    "clusters/{}/topics/{}/partitions/{:0>10}/records/{:0>20}.batch",
+                    self.cluster,
+                    topition.topic,
+                    topition.partition,
+                    floor - 1,
+                ));
+
+                self.object_store
+                    .list_with_offset(Some(&location), &start_after)
```
simon-engledew, line 1160: This uses list_with_offset to skip ahead past batches we don't care about.

tansu-storage/src/dynostore.rs
```diff
@@ -1255,15 +1359,41 @@ impl Storage for DynoStore {
...
-                .inspect_err(|error| error!(?error))
-                .map_err(|_| Error::Api(ErrorCode::UnknownServerError))?
+                _ => None,
+            };
+
+            let scan = matches!(offset_request, ListOffset::Timestamp(_))
+                || (offset_request == &ListOffset::Latest && stable.contains_key(topition));
+
+            while scan
+                && let Some(meta) = list_stream
```
simon-engledew, line 1395: I'm pretty sure this can also use the binary search. It makes the diff more complicated though 😬

simon-engledew: I had a stab at using binary search for every path here https://github.com/simon-engledew/tansu/compare/simon-engledew/minimal-binary-search...simon-engledew:tansu:simon-engledew/minimal-binary-search-all-paths

simon-engledew: After that, this change stops doing more work than necessary when fetching batches of records: https://github.com/simon-engledew/tansu/compare/simon-engledew/minimal-binary-search-all-paths...simon-engledew:tansu:simon-engledew/max-bytes

shortishly:
Hi @simon-engledew,

Thanks very much for keeping the surface area of your change nice and small ❤️, and your Rust look great to me  👍

The [list_with_offset](https://docs.rs/object_store/0.14.0/object_store/trait.ObjectStore.html#method.list_with_offset) on object store, states that "order is not guaranteed". If metadata were to be returned out of order, how would that affect your binary search?

simon-engledew:
👋 Yes, it's order dependent.  😞

I've done some digging and in practice both the InMemory and AmazonS3 backends both guarantee an ordered output, ~but I'll see if I can figure out how to rewrite it to be order independent~.

I _think_ it's good enough to check the page for the lowest entry and have implemented that. Surely if `list_offsets` doesn't  at least return the first thing from the thing you are listing then it's functionally useless 😅

simon-engledew:
👋 Hey sorry, I've been away a bit.

I still have these compat-librdkafka tests failing. Do you think I should be able to get them passing or are they currently WIP and expected to fail?

tansu-storage/src/dynostore.rs
```diff
@@ -497,6 +497,94 @@ impl DynoStore {
...
+                .list_with_offset(Some(&location), &start_after)
+        } else {
+            self.object_store.list(Some(&location))
+        };
+
+        list_stream
+            // `ObjectStore::list_with_offset` does not guarantee the order of returned
+            // metadata, so this selects the least key across the whole listing rather
+            // than relying on the first item yielded.
+            .try_fold(None, |least: Option<ObjectMeta>, meta| async move {
```
solace-aross, line 531: This looks like it solves the correctness question Peter raised, but I think it may undo the performance win the PR is aiming for.

`list_with_offset` returns an unbounded stream, and since order isn't guaranteed, this `try_fold` can never stop early - it has to consume every remaining object in the partition to be sure it found the true minimum. So each probe in the `batch_base_at_or_before` binary search costs O(remaining partition size) rather than O(1), and the outer ~log2(N) probes add up to on the order of N object reads (and several separate S3 round trips) instead of one full scan.

For the in-memory and S3 backends specifically - the ones this PR targets, and where the listing does come back ordered in practice per your testing - that ordering is never used to short-circuit. The fold still walks the whole remaining tail every time. So on the happy path, this may end up doing more work than the full scan it's replacing, not less.

Would a fast path be worth adding - trust the order when consecutive items come back non-decreasing, and only fall back to the full fold if that's ever violated? Might also be worth checking a request/item count against the `large_partition` test you added, to see how many objects each probe is actually touching.

solace-aross (commented):
Hi Simon, thanks for this - I really appreciate you keeping the diff small and adding test coverage for something this tricky.

I'm still new to contributing here and finding my way around the storage layer, so take this as an extra pair of eyes rather than the last word. One inline note below on the latest commit (`de1235e`), which answers Peter's ordering question above but may trade away some of the performance win this PR is going for. Nothing here feels blocking, just flagging it while it's fresh.

I wasn't able to confirm this with an actual run - the workspace here was mid-build on something else and I didn't want to sit on this while it finished - so the note below is from reading the code, not from measured numbers. Happy to be corrected.

tansu-storage/src/dynostore.rs
simon-engledew: Possibly? It's been so long since I've looked at this I've completely unloaded all the context from my head. 😅 

The actual hot-fix we tried running is a [much bigger change than this](https://github.com/simon-engledew/tansu/compare/main...simon-engledew:tansu:simon-engledew/concurrent) but breaking it up and verifying each stage's correctness is going to be more work than I can manage at the moment, sorry. (A colleague drafted the original fix with Claude and then I tried to reimplement it to understand it properly - all that is gone from my memory now 😂🙃 ) I think it would probably be best if I close this PR – after two months I honestly can't remember what I was going to do next and this was only ever supposed to be a first step in a series of about 7 PRs. 😂 

We never progressed to using running nisshi/tansu in production because this seemed like such a critical bug (– it became unusably slow after 3 weeks of messages) and I didn't trust my patch to be correct. 😅 It's a shame because I really love the concept! Good luck - I hope the branch I linked is useful!

Issue #694, opened 2026-07-11
SCRAM enforcement


HarrisonCentner:
## Overview

The broker treats a SASL exchange as authenticated whenever the underlying `rsasl` session reaches `State::Finished`. But `rsasl`'s SCRAM server reaches `State::Finished` even when the client's proof does not verify. The server does emit an error token, but a malicious client can ignore this. 

## Codepath

`is_authenticated()`  https://github.com/tansu-io/tansu/blob/510f3e2bd367fd19e6c9473aea02ec9bc30df085/tansu-auth/src/lib.rs#L106-L112

`Stage::Finished(Option<Success>)` — identity is optional  https://github.com/tansu-io/tansu/blob/510f3e2bd367fd19e6c9473aea02ec9bc30df085/tansu-auth/src/lib.rs#L85-L89

`SaslAuthenticateService` stores Finished(success) on any `State::Finished` and returns `ErrorCode::None`  https://github.com/tansu-io/tansu/blob/510f3e2bd367fd19e6c9473aea02ec9bc30df085/tansu-auth/src/authenticate.rs#L90-L105

The frame router admits every API once authenticated  https://github.com/tansu-io/tansu/blob/510f3e2bd367fd19e6c9473aea02ec9bc30df085/tansu-service/src/frame.rs#L263-L270 and https://github.com/tansu-io/tansu/blob/510f3e2bd367fd19e6c9473aea02ec9bc30df085/tansu-service/src/frame.rs#L325-L327

## Suggested fix

Require `Stage::Finished(Some(_))` in `is_authenticated()`, and have `SaslAuthenticateService` return `SaslAuthenticationFailed` when the exchange ends in `Stage::Finished(None)`.

HarrisonCentner:
Resolved by #696

PR #696, opened 2026-07-11
fix(auth): require a validated identity to treat SASL as authenticated

HarrisonCentner:
Fixes issue #694 

Adds an enforcement test that asserts a bad connection is rejected with `NotAuthenticated`.


HarrisonCentner:
Not sure if the new CI test is wanted. I'm happy to drop it.

shortishly:
Thank you for the PR!

PR #697, opened 2026-07-11
fix(auth): refuse SASL PLAIN instead of accepting any credentials

### LINKED ISSUE #695: Plain SASL authentication may be reacheable due to Cargo feature unification and plain passwords are not checked
## SASL `PLAIN` accepts any credentials: the callback checks that credentials are *present*, not *correct*

The broker's SASL `PLAIN` callback does not verify password.

## Reachability

In the default build `PLAIN` is disabled since `rsasl` is pinned without the `plain` feature. A `SaslHandshake("PLAIN")` fails with `NoSharedMechanism` before the callback ever runs.

However, Cargo unifies features across the entire dependency graph. So if another crate in the build pins `rsasl` with the `plain` feature, `plain` is enabled for tansu's `rsasl` too.

## The vulnerability                                                                                  
Once `PLAIN` is reachable, the callback returns `Success` as soon as the `AuthId` and `Password` properties are *present*. It never checks the password.
 
https://github.com/tansu-io/tansu/blob/510f3e2bd367fd19e6c9473aea02ec9bc30df085/tansu-auth/src/lib.rs#L182-L201

## Suggested fix

The implementation should refuse `PLAIN` in `check()`.

HarrisonCentner:
Fixes #695

### DIFF
--- a/tansu-auth/src/lib.rs
+++ b/tansu-auth/src/lib.rs
@@ -17,7 +17,7 @@ use rsasl::{
     config::SASLConfig,
     mechanisms::scram::properties::ScramStoredPassword,
     prelude::{SASLError, SASLServer, Session, SessionError, Validation},
-    property::{AuthId, AuthzId, Password},
+    property::{AuthId, AuthzId},
     validate::{Validate, ValidationError},
 };
 use std::{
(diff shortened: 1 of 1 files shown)


shortishly:
Thank you for the PR!

solace-aross (approved):
Thanks for this one, it's a real fix for a real hole. I'm still fairly new to contributing here and learning the codebase, so take my read with that in mind, but I dug into it carefully because it's auth code.

What I checked:
- Confirmed the vulnerability against current `main`: `nisshi-auth/src/lib.rs`'s `check()` still has the exact PLAIN-mechanism code path you found (the file moved from `tansu-auth` to `nisshi-auth` in the tansu->nisshi rename, but the logic is untouched). Before this fix, presenting any `AuthId`/`Password` pair returns `Success` without ever touching storage or verifying the password, matching issue #695 exactly.
- Verified the fix applies clean via a merge onto current `main` (git's rename tracking carries it across the `tansu-auth` -> `nisshi-auth` move without conflict), and that the patched `check()` now returns `AuthError::UnknownMechanism` for PLAIN unconditionally, so the SCRAM-only branch is the only path that can produce `Success`.
- Checked whether this trades one bug for a worse one: the `else` branch already returns `UnknownMechanism` for any mechanism that isn't PLAIN or SCRAM-*, so refusing PLAIN there is consistent with how every other unsupported mechanism is already handled, not a new code path.
- Checked the CI failures on this PR (compat-franz-go / compat-librdkafka failing on postgres/s3/sqlite): those are failing the same way on `main`'s last three runs, so they're pre-existing and unrelated to this change, not something this PR needs to fix.

One thing I'd love your take on, not a blocker: there's no test exercising the PLAIN path (before or after), so nothing will flag it if this ever regresses, e.g. in a future refactor of `check()`. That said, the SCRAM branch right next to it has no dedicated unit test either, so this looks like an existing gap in this module rather than something introduced here. Worth a follow-up issue if nobody's tracking it already, but I don't think it should hold up this fix.

Approving. Solid, minimal, well-targeted change, and a good catch on the underlying issue.

PR #715, opened 2026-08-12
fix(storage): keep record headers in produce order

OrHayat:
Kafka record headers are an ordered list that may repeat a key.
The header table keyed on (topition, offset_id, k) and header_fetch.sql had no order by, so postgres served headers from that primary key index in key order rather than produce order, and a repeated key could not be stored at all.

Caught by librdkafka 0073_headers and 0085_headers against postgres. 0073 produces msgid, static, multi, multi, multi, null, empty and validates by index; sorted by key that starts with "empty", which is what it reported at index 0. 0085 produces header_0..header_260 and got header_10 back at index 2, the byte-sorted position.

The existing round-trip coverage in nisshi-broker/tests/fetch.rs compares headers as a BTreeSet, so it could not see either ordering or duplicates -- only the compat suite caught this.

Add an ordinal column recording the position of each header within its record, key the table on (topition, offset_id, ordinal) so repeated keys are storable, and order the fetch by it. The DDL and SQL are shared, so this covers the pg, lite and limbo backends.

  The header table changes shape and the project has no migration framework: an existing database needs to be recreated, or migrated by hand:
```sql
  alter table header add column ordinal int;
  update header h set ordinal = n.rn - 1
  from (select topition, offset_id, k,
               row_number() over (partition by topition, offset_id order by k) as rn
        from header) n
  where h.topition = n.topition and h.offset_id = n.offset_id and h.k = n.k;
  alter table header alter column ordinal set not null;
  alter table header drop constraint header_pkey;
  alter table header add primary key (topition, offset_id, ordinal);
```

### DIFF
--- a/etc/initdb.d/010-schema.sql
+++ b/etc/initdb.d/010-schema.sql
@@ -156,9 +156,10 @@ order by
 create table if not exists header (
     topition int,
     offset_id bigint,
-    k bytea,
-    primary key (topition, offset_id, k),
+    ordinal int,
+    primary key (topition, offset_id, ordinal),
     foreign key (topition, offset_id) references record (topition, offset_id) on delete cascade,
+    k bytea,
     v bytea,
     last_updated timestamp default current_timestamp not null,
     created_at timestamp default current_timestamp not null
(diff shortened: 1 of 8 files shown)


OrHayat:
rebased

OrHayat:
rebased

PR #716, opened 2026-08-12
feat(storage): report aborted transactions to consumers in postgres

OrHayat:
aborted_transactions was a stub returning an empty list, so a read_committed consumer was never told which offset ranges were aborted and client libraries delivered aborted data as if committed. 
The first commit implements it: find each producer's abort markers, compute each  aborted range's first offset, and wire the result into the Fetch response, kept cheap by a small partial index over control-batch records.

  1. A retried TxnOffsetCommit hit the staging tables' unique constraints and failed the whole request. Kafka requires the latest staged offset to win within the transaction.
  2. An offsets-only transaction (sendOffsetsToTransaction with nothing produced) never reset the transaction state, so EndTxn silently dropped its staged offsets and leaked them into the producer's next transaction.
  3.  AddOffsetsToTxn now begins the transaction the same way AddPartitionsToTxn does, which also lets the timeout sweep cover offsets-only transactions.
  4. TxnOffsetCommit validated nothing before staging: a sweep-fenced producer's stale epoch, staging into an already-finalized transaction,  and a nonexistent topic or partition were all acked as success. Validation now runs before any insert and returns ProducerFenced,  InvalidTxnState and UnknownTopicOrPartition respectively.
  5. Two debug views joined producer_epoch on itself, matching rows by coincidence of id sequences.

### DIFF
--- a/nisshi-storage/src/pg.rs
+++ b/nisshi-storage/src/pg.rs
@@ -7430,4 +7430,141 @@ mod tests {
 
         Ok(())
     }
+
+    /// Kafka permits TxnOffsetCommit to be sent repeatedly within one transaction -- a
+    /// retry after a lost response, or a later sendOffsetsToTransaction for the same
+    /// partition -- and the latest committed_offset must win. The staging inserts have no
+    /// conflict handling, so the second commit for the same (transaction, consumer group)
+    /// violates txn_offset_commit's unique constraint and fails the whole request.
+    #[tokio::test]
+    async fn txn_offset_commit_second_stage_overwrites_first() -> Result<()> {
+        use nisshi_sans_io::txn_offset_commit_request::{
+            TxnOffsetCommitRequestPartition, TxnOffsetCommitRequestTopic,
+        };
+
+        let cluster = alphanumeric_string(15);
+
+        let storage = Postgres::builder(CONNECTION)?
+            .cluster(cluster.as_str())
+            .node(rng().random_range(0..i32::MAX))
+            .build();
+
+        if let Err(err) = storage.connection().await {
+            eprintln!("skipping txn_offset_commit_second_stage_overwrites_first: {err:?}");
+            return Ok(());
+        }
+
+        storage
+            .register_broker(BrokerRegistrationRequest {
+                broker_id: 111,
+                cluster_id: cluster.clone(),
+                incarnation_id: Uuid::now_v7(),
+                rack: None,
+            })
+            .await?;
+
+        let topic_name = alphanumeric_string(15);
+
+        _ = storage
+            .create_topic(
+                CreatableTopic::default()
+                    .name(topic_name.clone())
+                    .num_partitions(1)
+                    .replication_factor(0)
+                    .assignments(Some([].into()))
+                    .configs(Some([].into())),
+                false,
+            )
+            .await?;
+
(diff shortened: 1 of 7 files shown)


nisshi-storage/src/pg.rs
```diff
@@ -3702,30 +3783,101 @@ impl Storage for Postgres {
...
+                        .map(TxnState::try_from)
+                        .transpose()?;
 
-            (producer_id, epoch)
-        } else {
-            (None, None)
+                    if producer_id != offsets.producer_id {
+                        Some(ErrorCode::UnknownProducerId)
+                    } else if producer_epoch != offsets.producer_epoch
+                        || status != Some(TxnState::Begin)
```
solace-aross, line 3840: This new check requires `txn_detail.status == 'BEGIN'` before a `TxnOffsetCommit` can succeed. That status only gets set to `'BEGIN'` by `txn_add_partitions` or by this PR's new call to `txn_detail_update_started_at.sql` inside `txn_add_offsets` (line ~3666). But `init_producer` (`bump_or_create_producer` -> `txn_detail_insert.sql`) creates the `txn_detail` row with `status = NULL` and nothing else sets it.

The existing shared tests `simple_txn_commit_offset_commit` and `simple_txn_commit_offset_abort` (`nisshi-broker/tests/txn.rs:130` and `:268`) call `init_producer` then go straight to `txn_offset_commit`, without `AddPartitionsToTxn` or `AddOffsetsToTxn` in between. With `status` still `NULL`, this check now returns `InvalidTxnState` where the tests expect `ErrorCode::None`. Confirmed on this PR's own CI: both tests fail on `postgres:16/17/18` with `left: None, right: InvalidTxnState`, while the `memory`, `lite`, and `slatedb` variants of the same tests still pass, which isolates it to this change.

Given the commit message's own framing (`AddOffsetsToTxn now begins the transaction the same way AddPartitionsToTxn does`), the stricter check does look like the intended, more-correct behavior; a real Kafka client would send `AddOffsetsToTxn` before an offsets-only commit. If that's right, the two shared tests need an `AddOffsetsToTxn`/`txn_add_offsets` call added before their `txn_offset_commit`. If instead this validation is meant to be more lenient (e.g. tolerate a fresh, never-begun transaction the way a bare offsets-only commit apparently could before), the check itself needs to change. Either way, right now the PR fails its own pre-existing test suite, so I don't think this can merge as-is.

solace-aross (changes requested):
I'm still new to this codebase and learning as I go, so please take this in that spirit, but I wanted to flag something I found while working through the transaction path.

The fixes in this PR are careful and the write-up in `record_control_marker_select.sql` (the lag()-over-producer_id reasoning, with benchmarked alternatives) is genuinely excellent documentation. Thank you for that.

One thing stands in the way of merging as-is: the new `TxnOffsetCommit` state check breaks two of the project's own existing tests on postgres, and it's already red in this PR's CI (run 33414039141, `test (ubuntu-latest, postgres:16/17/18)`, 861 passed / 2 failed). See the inline comment for the mechanism. I don't think this needs new investigation on your end, just a decision on which side to fix (see the note in the inline comment) and a test update either way.

Everything else in the diff looks solid to me. Happy to take another look once the tests are green.

PR #720, opened 2026-09-16
ci: gate package/smoke/release on repo-scoped PRs instead of a named actor

sgamelin:
## Summary

The `package` image push and the `smoke` and `release` jobs in `.github/workflows/ci.yml` were gated on a single hard-coded GitHub username. This replaces those three checks with a condition that trusts anyone who can push to this repository, so the workflow no longer needs editing when maintainers change.

```yaml
${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository }}
```

- **`push` events** (`main`, `v*` tags): only users with write access can push, so the event itself is already a trusted-actor check.
- **`pull_request` events from a branch in this repository**: only users with write access can push a branch here, so these PRs continue to push the image and run smoke tests.
- **Pull requests from forks**: the head repository differs from `github.repository`, so they continue to skip the gated steps. This also lines up with when secrets are available to the run.

No API call or extra token is needed. Job dependencies and build flow are unchanged; only the conditions differ.

## Why not `author_association`

An org-membership check (`author_association` in `OWNER`/`MEMBER`) was tried first and rejected. For members whose org membership is private, the Actions event payload reports `COLLABORATOR` or `CONTRIBUTOR` rather than `MEMBER`, so the gate silently fails for them. The repo-scoped check does not depend on membership visibility.


sgamelin:
**Verification**

The gate expression was exercised in isolation with a stub `ci.yml` (two jobs, no build) on two throwaway draft PRs, now closed. The stub ran the condition as a job-level `if:` and also printed the raw values it depends on.

| PR | Head repo | `head.repo.full_name == github.repository` | Gated job |
|---|---|---|---|
| #721, branch in this repo ([run](https://github.com/nisshi-io/nisshi/actions/runs/35141414198)) | `nisshi-io/nisshi` | `true` | ran |
| #722, from a fork ([run](https://github.com/nisshi-io/nisshi/actions/runs/35141456950)) | `sgamelin/nisshi` | `false` | skipped |

The same runs also showed why the `author_association` approach was dropped: with private org membership the payload reported `COLLABORATOR`, and the membership gate evaluated `false` in both cases.

On this PR's own [run](https://github.com/nisshi-io/nisshi/actions/runs/35141659213), `package` should run with `push: true`, `smoke` should run rather than skip, and `release` should be scheduled (its upload step is still tag-only).

PR #723, opened 2026-09-16
fix(service): bound request frames read from unauthenticated peers

solace-aross:
The request frame-size guard never worked, and no listener turned it on. An unauthenticated client can send a 4-byte length prefix of `0x7FFFFFFF` and make the broker or proxy allocate a 2 GiB buffer per connection before SASL runs. A negative length prefix panics the connection task instead.

Three defects, all on the same trust boundary (the Kafka length prefix read from the wire):

- `TcpBytesService::wait` (`nisshi-service/src/stream.rs`) had the comparison inverted. With a limit set it rejected every frame smaller than the limit and admitted every frame larger than it.
- `frame_length` (`nisshi-service/src/lib.rs`, duplicated in `nisshi-client/src/lib.rs`) did `i32 as usize + 4`. A negative size sign-extends and overflows; in release it wraps into a `vec!` capacity-overflow panic.
- `TcpContext::default()` left `maximum_frame_size` as `None`, and both listeners (`nisshi-broker/src/service.rs`, `nisshi-proxy/src/lib.rs`) used the default. Production ran with no cap at all.

Fix:

- `frame_length` validates the prefix and returns `Result`, rejecting negative sizes with a new `Error::InvalidFrameLength(i32)`. It is now `pub`, and `nisshi-client` uses it instead of its own copy, so there is one validator. Nothing in the workspace matches on `nisshi_service::Error` exhaustively.
- `wait` rejects when the declared payload exceeds the limit. The bound and the reported `FrameTooBig(n)` are on the payload, the same quantity Kafka's `socket.request.max.bytes` bounds.
- `TcpContext::default()` now bounds payloads at 100 MiB (`DEFAULT_MAXIMUM_FRAME_SIZE`, the Kafka default). Broker and proxy pick this up without changes; `.maximum_frame_size(None)` opts out. This is the one behaviour change for legitimate clients. The Kafka client default `max.request.size` is 1 MiB, and nothing under `compat/` sends frames near 100 MiB.

Behaviour change worth a release note: `nisshi-service` is published, and any downstream listener built on `TcpContext::default()` goes from unbounded to 100 MiB. Frames above that now close the socket with `FrameTooBig` in the log instead of being read. `nisshi_service::Error` also gains a variant.

What this does not do, for follow-up:

- The full declared size is still allocated before the body arrives, and there is no read-idle timeout or connection bound. 4 bytes now buys 100 MiB per connection instead of 2 GiB; incremental reads and a pre-auth idle timeout would close the rest.
- The limit is a constant, not a CLI flag. Kafka makes this configurable; `EnvVarExp<T>` is the obvious vehicle if an operator needs it.
- `maximum_frame_size(None)` is now an explicit opt-out rather than "take the default". The doc says so, but the `Option` signature does not; a `usize` setter plus an explicit `unbounded()` would make the next unbounded listener a visible decision. Left out here to keep this PR to the fix.

Verified:

- `cargo test -p nisshi-service --all-targets` green at HEAD (baseline) and with the change (6 new tests plus the existing 8).
- Red-green: 4 of the new tests fail at HEAD (`frame_within_limit_is_accepted`, `frame_over_limit_is_rejected`, `negative_frame_length_is_rejected`, `serve_rejects_oversized_frame_without_reading_body`). Reverting the fix reproduces the red, including the arithmetic overflow panic in `frame_length`.
- `cargo clippy --workspace --all-features --all-targets -- -D warnings` clean.
- `cargo fmt --all --check` clean.
- Reviewed by an independent principal-architect pass; its two in-scope findings (proxy listener uncapped, `nisshi-client` duplicate) are folded in above.


nisshi-service/src/stream.rs
```diff
@@ -34,8 +34,16 @@ use tracing::{debug, error, instrument};
...
     BYTES_RECEIVED, BYTES_SENT, Error, REQUEST_DURATION, REQUEST_SIZE, RESPONSE_SIZE, frame_length,
+    frame_size,
 };
 
+/// The largest request payload a listener accepts unless [`TcpContext::maximum_frame_size`]
+/// says otherwise, matching the Apache Kafka default for `socket.request.max.bytes`.
+///
+/// The size prefix is read before authentication, so an unbounded listener lets
+/// a client make it allocate up to 2 GiB per connection by sending 4 bytes.
+pub const DEFAULT_MAXIMUM_FRAME_SIZE: usize = 100 * 1024 * 1024;
```
sgamelin, line 45: Non-blocking. The existing docs in this file are single-line with no trailing period (`/// A [`Layer`] that listens for TCP connections`). This one is two paragraphs with periods, and the second paragraph repeats the threat rationale from the PR description. Consider trimming to the first sentence so it reads like its neighbours. Same applies to the `maximum_frame_size` setter doc at line 162 and the `Echo` helper at line 490, which is the only documented test helper in the workspace.

nisshi-service/src/lib.rs
```diff
@@ -316,8 +317,22 @@ impl<T> From<PoisonError<T>> for Error {
...
+fn frame_size(encoded: [u8; 4]) -> Result<usize, Error> {
+    let size = i32::from_be_bytes(encoded);
+
+    usize::try_from(size).map_err(|_| Error::InvalidFrameLength(size))
+}
+
+/// Total length of a frame, the 4 byte size prefix plus the payload it declares.
+///
+/// Fails with [`Error::InvalidFrameLength`] when the prefix declares a negative size.
+pub fn frame_length(encoded: [u8; 4]) -> Result<usize, Error> {
```
sgamelin, line 334: Non-blocking, same style note. The second paragraph on `frame_size` restates what the code below it does (`usize::try_from` fails on negative). The `frame_length` doc pointing at `Error::InvalidFrameLength` is worth keeping since it is `pub`; the private `frame_size` could drop to one line.

sgamelin (approved):
Approving. The three defects are real and the fix closes each one at the right spot: `frame_size` validates the untrusted prefix once, `wait` bounds the payload rather than the framed length, and `TcpContext::default()` gives every listener the Kafka default. The duplex test is the one that proves the check happens before the allocation, which is the property that matters here. Thanks for keeping the scope tight and listing the follow-ups.

Cross-checked the Kafka references against upstream trunk: `socket.request.max.bytes` defaults to `100 * 1024 * 1024` (`SocketServerConfigs.java`), producer `max.request.size` defaults to `1024 * 1024` (`ProducerConfig.java`), and `NetworkReceive.java` rejects a negative prefix and compares the declared payload size against the limit, so bounding the payload here matches Kafka's own semantics.

The compat CI failures are pre-existing: the same four jobs fail on main at 4a53d40 because the broker cannot reach postgres in that job. Not related to this change.

Two non-blocking notes on doc comment style inline. Take or leave them.

PR #724, opened 2026-09-16
Rama zero dot four

shortishly:


solace-aross:
Not mergeable as-is. This branch predates #723 ("bound request frames read from unauthenticated peers", merged ~17h before this review) and was never rebased — `gh` reports the PR as `CONFLICTING`. That's the root cause of the two blocking issues below. The refactor itself looks sound; the risk is in how the merge conflict gets resolved.

**Blocking**

1. Frame-size guard is inverted, and re-opens the pre-#723 hole.

   `nisshi-service/src/stream.rs:399`:
   ```rust
   if maximum_frame_size.is_some_and(|maximum_frame_size| maximum_frame_size > frame_length(size)) {
       return Err(Into::into(Error::FrameTooBig(frame_length(size))));
   }
   ```
   This rejects small frames and admits oversized ones. `frame_length` (`nisshi-service/src/lib.rs:338`) has no bounds check: `i32::from_be_bytes(encoded) as usize + 4`. A negative length prefix sign-extends to `usize::MAX`, then overflows or wraps and panics on the slice. A 4-byte write from an unauthenticated peer kills the connection task.

   On top of that: `TcpContext`'s derived `Default` (`stream.rs:136`) gives `maximum_frame_size: None`, unbounded by default, where main hand-writes `Some(100 MiB)`. And the four regression tests #723 added aren't here — they're written against the `Context<State>` API this PR removes, so this isn't a mechanical rebase. The guard and its tests need to be re-ported into the new `Extensions`-based `wait`/`read`.

2. New `ReadHalfService` has no size check at all.

   `nisshi-service/src/stream.rs:253`: `vec![0u8; frame_length(size)]`, straight off the wire, no limit. Not wired into broker/client/proxy yet (only `nisshi-service/tests/tcp.rs` uses it), so not live today, but it needs `maximum_frame_size` before anything connects to it.

3. Storage OTel metrics are gone, dashboard still queries them.

   Every `u64_counter`/`u64_histogram` in the storage layer is removed and not replaced (`STORAGE_CONTAINER_REQUESTS`/`_ERRORS` etc.). `etc/grafana/dashboards/home.json:6013` and `:6074` still query `nisshi_storage_container_requests_total`, and that file isn't touched by this PR. Two panels go blank on deploy, silently.

**Fix in the same pass**

- `nisshi-client/src/lib.rs:799` carries its own copy of the unguarded `frame_length` that main already deduplicated by importing from `nisshi-service`. Lower risk (needs a malicious broker/proxy) but same defect.
- `nisshi-storage/src/service/produce.rs:192`: `ProduceService` went from a zero-field unit struct to `ProduceService<G> { storage: G }`, and `Storage: Debug` is a supertrait, so `debug!(?self, ?error_code)` now Debug-dumps the whole backing store on every per-record error. Same shape recurs across `nisshi-storage/src/service/*` — worth an `#[instrument(skip(self))]` sweep rather than a one-off fix.
- `nisshi-service/src/stream.rs:212`: `TcpStreamService::serve` uses `.fork()` on request extensions where this PR's own new CLAUDE.md convention calls for `.clone()` on same-session requests (so state like `AuthenticationExtension` survives a multi-round-trip SCRAM flow). Also unwired today, so a trap for whoever connects it next, not a live break.

**Follow-up, not blocking**

- Storage scheme dispatch moved from exact match to unanchored regex substring match (`r"postgres|postgresql"`, `r"s3"`, etc.). No collision among today's 8 schemes, but `postgress://` now silently matches Postgres instead of erroring cleanly. Anchor the patterns.
- The 8-call factory registration block is duplicated verbatim in three places: `broker.rs:672`, `administrator.rs:3191` (test helper), and `tests/common/mod.rs:93` (which is already missing one factory). Worth a shared builder.

**Non-issue**

The commented-out `produce.rs` unit tests aren't a coverage gap. Equivalent, broader tests already live in `nisshi-broker/tests/produce.rs`, running against InMemory, Lite, Postgres, and SlateDb instead of just in-memory. Fine to just delete the ~480 commented-out lines rather than carry them.

Can you rebase onto main and re-resolve the `stream.rs`/`lib.rs` conflict specifically? That's where #723's fix either survives or doesn't.

solace-aross (changes requested):
Formalizing the request-changes verdict from my prior comment above (unchanged since — no new commits, still `CONFLICTING` against main). See that comment for the three blocking issues (inverted frame-size guard reopening the pre-#723 hole, unguarded `ReadHalfService`, dropped storage OTel metrics still queried by the Grafana dashboard) and the fix-in-the-same-pass items. Needs a rebase onto main with the `stream.rs`/`lib.rs` conflict re-resolved before this can move forward.

solace-aross (changes requested):
Two of the three blockers from my last review are fixed and verified. One is still open.

**Resolved**

1. Frame-size guard direction. `TcpContext::default()` now sets `maximum_frame_size: Some(DEFAULT_MAXIMUM_FRAME_SIZE)` instead of the derived `None` (`nisshi-service/src/stream.rs:150-157`), and `TcpBytesService::wait`'s comparison (`stream.rs:441`) is the correct direction: `frame_size > maximum_frame_size`. This is the path actually wired to the broker (`nisshi-broker/src/service.rs:32-33`, `TcpContextService<TcpBytesService<...>>`), and it's covered by `listeners_are_bounded_by_default` plus the ported `frame_over_limit_is_rejected`/`negative_frame_length_is_rejected` tests. `frame_size()` in `lib.rs` also fails clean on a negative prefix now (`usize::try_from`), so the panic risk is gone at the source, not just patched at one call site.

2. `ReadHalfService` unbounded allocation — fixed but the regression test doesn't prove it. `4f9e750` adds the right guard (`stream.rs:287-296`, correct comparison, runs before the `vec![0u8; frame_length(size)?]` alloc). But the commit message says this is covered by re-enabling `serve_rejects_oversized_frame_without_reading_body`; it isn't. That test calls `service()`, which builds a `TcpBytesService` (line 633), not a `TcpStreamService`/`ReadHalfService`. I deleted the entire new guard block and reran the full `nisshi-service` suite: all 14 tests still passed, including that one — it's exercising `TcpBytesService::wait`'s independent check, never touching `ReadHalfService`. Not blocking, since `ReadHalfService` still has exactly one caller (`TcpStreamService`, only used in `tests/tcp.rs`) and no production path reaches it yet. But whoever wires it up next will have a false green light. Worth a test that actually builds a `TcpStreamService` and sends it an oversized frame before this lands anywhere live.

**Still open**

3. Storage OTel metrics. `STORAGE_CONTAINER_REQUESTS`/`STORAGE_CONTAINER_ERRORS` are still gone from `nisshi-storage` at this head (confirmed: zero matches for either name, and no `u64_counter`/`u64_histogram` left in `nisshi-storage/src/lib.rs` or `service/*.rs`). `etc/grafana/dashboards/home.json` still queries `nisshi_storage_container_requests_total` in two panels. Neither this branch nor anything picked up in the rebase restores or replaces them, so those panels go blank on deploy. This is the one still blocking.

Rebase is clean now (`gh` reports `MERGEABLE`), so no merge-conflict risk left to resolve. Once the storage metrics are back (or intentionally replaced with an equivalent), I'm good to approve.

shortishly:
> 1. Frame-size guard is inverted, and re-opens the pre-[fix(service): bound request frames read from unauthenticated peers #723](https://github.com/nisshi-io/nisshi/pull/723) hole.
>    `nisshi-service/src/stream.rs:399`:
>    ```rust
>    if maximum_frame_size.is_some_and(|maximum_frame_size| maximum_frame_size > frame_length(size)) {
>        return Err(Into::into(Error::FrameTooBig(frame_length(size))));
>    }
>    ```

I think the frame size guard is back in place.

>    On top of that: `TcpContext`'s derived `Default` (`stream.rs:136`) gives `maximum_frame_size: None`, unbounded by default, where main hand-writes `Some(100 MiB)`. And the four regression tests [fix(service): bound request frames read from unauthenticated peers #723](https://github.com/nisshi-io/nisshi/pull/723) added aren't here — they're written against the `Context<State>` API this PR removes, so this isn't a mechanical rebase. The guard and its tests need to be re-ported into the new `Extensions`-based `wait`/`read`.

Tests have been updated to use Extensions.

Note that I plan to remove `TcpContext` at some point: It kind of made sense when Rama required a context for each serve function. With the new extensions mechanism, I think `TcpContext` should become redundant. I am looking at using `ReadHalfService` and `WriteHalfService`:

- On the client use `WriteHalfService` to write the request and `ReadHalfService` to read the responses
- On the server (broker/proxy) use `ReadHalfService` to read the request and `WriteHalfService` to write the response
- Use extensions on the Input to enable other features (e.g., max wait times/etc)


Also `TcpListenerLayer` might make more sense as a [TcpListener](https://ramaproxy.org/docs/rama/tcp/server/struct.TcpListener.html)


> 2. New `ReadHalfService` has no size check at all.
>    `nisshi-service/src/stream.rs:253`: `vec![0u8; frame_length(size)]`, straight off the wire, no limit. Not wired into broker/client/proxy yet (only `nisshi-service/tests/tcp.rs` uses it), so not live today, but it needs `maximum_frame_size` before anything connects to it.

Size check is now incorporated.

> 3. Storage OTel metrics are gone, dashboard still queries them.
>    Every `u64_counter`/`u64_histogram` in the storage layer is removed and not replaced (`STORAGE_CONTAINER_REQUESTS`/`_ERRORS` etc.). `etc/grafana/dashboards/home.json:6013` and `:6074` still query `nisshi_storage_container_requests_total`, and that file isn't touched by this PR. Two panels go blank on deploy, silently.

StorageContainer used to delegate to storage. Some metrics were being generated by the container itself.

StorageContainer is now just a factory for Storage, individual engines will use their own metrics rather than through the container.

PR #726, opened 2026-09-17
ci: fast dependency and workflow-security gates, pin all actions to SHA

solace-aross:
## Summary

First step toward a faster, safer CI foundation, adapted from patterns proven on a sibling Go project without pulling in its enterprise-specific compliance tooling.

- **New: `dependencies.yml`** — `cargo-deny` over the full dependency graph (RustSec advisories, license compatibility, banned crates, trusted sources).
- **New: `workflow-lint.yml`** — `actionlint` + `zizmor` over the workflow files themselves (no `paths:` filter — a required check that's path-filtered can permanently block a PR that never fires it).
- **Every third-party Action pinned to a commit SHA** instead of a mutable tag (54 references across `ci.yml`/`release.yml`), closing a real supply-chain gap zizmor flagged.
- **Caching re-enabled** on `check`/`fmt`/`clippy` (and `cargo-publish-dry-run`/`third-party-license` for consistency) — `cache: false` dated back to a Feb 2025 workaround for an unrelated warning, not a deliberate call. Gated with `cache-save-if: push-only` for cache-budget hygiene.
- **4 shellcheck quoting bugs fixed**, `just-version` pinned on all `setup-just` calls.
- **2 real RUSTSEC advisories (h2, rustls) resolved directly** via a semver-compatible `Cargo.lock` bump rather than waived. 4 others (hickory-proto x2, quick-xml x2) are genuinely blocked upstream and tracked in `deny.toml` with dated, per-ID reasons.

## What this deliberately does not do

- Does not touch the existing `test`/`build-storage`/`build-storage-lake`/`compat-*` job matrix or its `needs:` graph — this is additive, not a restructure.
- Does not change branch protection. **Nothing here is required to merge yet** — that needs repo admin access, which I don't have. Follow-up ask incoming separately.
- No DCO/signed-commit/CODEOWNERS — deferred to a later iteration once this baseline is proven.

## Test plan

- [x] `cargo check --workspace --all-features --all-targets` — clean (826 crates)
- [x] `actionlint .github/workflows/*.yml` — clean
- [x] `zizmor --min-confidence medium --min-severity medium .github/workflows/` — clean
- [x] `cargo-deny --all-features --locked check --warn unmaintained` — clean (`advisories ok, bans ok, licenses ok, sources ok`)
- [x] Every pinned SHA resolved and verified against GitHub's API
- [x] Reviewed by an independent pass that caught and led to fixing 8 issues (advisory waivers that were actually fixable, a false "no toolchain needed" claim, an overclaimed security comment, a missing zizmor online-audit token, a stale tag comment, and more)


.github/workflows/workflow-lint.yml
```diff
@@ -0,0 +1,64 @@
...
+# WHY THERE IS NO `paths:` FILTER
+#
+# The obvious shape for this workflow is `pull_request` filtered to
+# `.github/workflows/**`. That shape is wrong for a *required* status check:
+# GitHub's branch protection waits for a check to report on the exact commit
+# being merged, and a path-filtered workflow never runs (and never reports)
+# on a PR that doesn't touch that path. A required check that never fires
+# blocks the PR forever rather than passing it. Running on every PR is the
+# price of using this as a required check; the job itself is seconds, not
+# minutes, so it's cheap either way.
```
sgamelin: The diagnosis here is right (GitHub's [troubleshooting docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks) confirm a workflow skipped by `paths:` leaves its checks pending), but "run everything on every PR" isn't the only option. The same docs note that a *job* skipped by an `if:` conditional reports success, which is what makes the usual pattern work:

- keep the workflow triggering on every PR (no `paths:`), so it always reports;
- add a small gate job using [`dorny/paths-filter`](https://github.com/dorny/paths-filter) to detect changed files;
- put `if: needs.changes.outputs.workflows == 'true'` on the real jobs.

When nothing relevant changed, the jobs skip in seconds and the required check still goes green. Not worth it for actionlint/zizmor since they're cheap, but it would matter if `dependencies.yml` (needs a toolchain + `cargo metadata`) later becomes required. Suggest mentioning that pattern in one line here so the next person doesn't reach for `paths:` again, and trimming the rest of this header down. The full reasoning already lives in the PR description.

.github/workflows/ci.yml
```diff
@@ -87,12 +89,19 @@ jobs:
...
+      - uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1
         with:
           components: clippy, rustfmt
-          cache: false
+          # Every PR restores the cache but doesn't write to it; only a push to
+          # main refreshes it. GitHub already scopes a pull_request run's cache
+          # writes to that PR's own ref, so this isn't a poisoning fix - it's
+          # hygiene: it keeps the 10GB repo cache budget from being consumed by
+          # every open PR's cache entry instead of one shared entry per branch.
+          cache-save-if: ${{ github.event_name == 'push' }}
```
sgamelin, line 107: This six-line comment is repeated verbatim on `check`, `fmt`, and `clippy`, and then absent on `cargo-publish-dry-run` and `third-party-license`, which use the same setting. Suggest writing it once (top of file, or a YAML anchor the jobs reference) and leaving the other sites bare. Duplicated comments drift the first time one copy gets edited.

.github/workflows/dependencies.yml
```diff
@@ -0,0 +1,40 @@
...
+# standalone binary; this job still installs the pinned binary directly
+# (rather than through `cargo install cargo-deny`) since that's much faster
+# than compiling it from source.
+#
+# `deny.toml` ignores a fixed, dated list of pre-existing RUSTSEC advisories
+# that have no upstream fix yet (see the comment there) and downgrades
+# `unmaintained` findings to non-blocking - being unmaintained isn't the same
+# as being vulnerable, and several stable, frozen micro-crates in this tree
+# are unmaintained by design, not by neglect. Any advisory not already on
+# that list still fails the build.
```
sgamelin: Same note as `workflow-lint.yml`: this header is useful "why" content, but at this length it's more likely to rot than be reread when the YAML below changes. A two- or three-line summary here plus the detail in the PR description (or a short `docs/ci.md`) would age better.

sgamelin (approved):
Looks good. SHA pinning, the shellcheck fixes, and restoring the cache that `e179c32` turned off as a warning workaround are all clear wins, and the RUSTSEC bumps being fixed rather than waived is the right call.

Two non-blocking notes inline: the paths-filter reasoning in `workflow-lint.yml` is correct but there is a standard way to skip jobs by path without deadlocking a required check (gate job + `dorny/paths-filter` + job-level `if:`), worth a pointer; and the workflow comments are duplicated in `ci.yml` and long in the two new files. Happy for both to land as a follow-up.

.github/workflows/workflow-lint.yml
solace-aross: Fair point, done in c87ebc9. Trimmed the header and added a one-line note pointing at `dorny/paths-filter` as the pattern to reach for if a heavier job needs it later.

.github/workflows/ci.yml
solace-aross: Agreed, done in c87ebc9. Moved the rationale to a single comment at the top of ci.yml and left `cache-save-if` bare on all five jobs that use it.

.github/workflows/dependencies.yml
solace-aross: Done in c87ebc9, trimmed the same way as workflow-lint.yml.

PR #727, opened 2026-09-17
fix(service,broker): stop accept-loop wedge on a transient accept() error

solace-aross:
Small one while doing a robustness pass on the network layer: found this in both accept loops, so flagging it as one PR rather than two.

## Problem

Both TCP accept loops in this codebase write the accept branch as
`Ok((stream, addr)) = req.accept() => { ... }` inside `tokio::select!`.
`tokio::select!` disables an arm whose head pattern doesn't match the
resolved value, for the rest of that macro invocation; the future is
already consumed, not retried within that call.

- **`nisshi-service/src/stream.rs`'s `TcpListenerService::serve`** (used by
  `nisshi-proxy` and `nisshi-client`): a single `accept()` error
  (`ECONNABORTED`, `EMFILE`, `ENFILE`, ...) disables that arm. If there are
  no in-flight connections at that moment, the `join_next` branch is also
  inactive (it's gated on a non-empty `JoinSet`), leaving only the
  cancellation branch active, so that `select!` call blocks until shutdown.
  One transient error can permanently stop that listener from accepting new
  connections.
- **`nisshi-broker/src/broker.rs`'s `Broker::listen`** (the actual broker
  binary): same root defect, but this loop also has two unconditional
  `interval.tick()` branches (maintenance sweep, default 10 min; txn sweep,
  default 10s), so the `select!` call is eventually forced to return by one
  of those, bounding the stall to that interval instead of forever. Same
  bug, lower severity.
- Same file, same arm: `stream.set_nodelay(true)?;` used `?`, so a single
  per-socket error there propagated out of `listen()` entirely and ended the
  whole accept loop. Same failure shape, different trigger.

## Fix

Bind the raw `Result` from `accept()` and `match` on it inside the arm
body instead of in the `select!` pattern, so the arm always resolves
regardless of `Ok`/`Err`. On `Err`, log and continue; back off (5ms
doubling, capped at 200ms) only when the error isn't a routine
`ConnectionAborted`, to avoid spinning the loop at 100% CPU under sustained
resource exhaustion. The backoff sleep is deliberately kept short, since it
blocks the whole `select!` call, including the `join_next` branch that's
what actually frees resources under fd pressure.

`stream.set_nodelay` errors in `broker.rs` are now handled locally (logged,
connection dropped) instead of propagating via `?`.

## Testing

`nisshi-service`'s `TcpListenerService` is now generic over a small
`Acceptor` trait, blanket-implemented for `TcpListener`. That needs no
change at any existing call site (`nisshi-proxy`, `nisshi-client`; both
still build, existing tests/doctests pass unchanged). A new unit test
(`stream::tests::accept_error_does_not_wedge_the_loop`) drives the service
against a scripted `Acceptor` returning `Err(ConnectionAborted)` then a
real accepted connection, and proves the connection after the error is
actually served (echoes a byte over the raw stream) rather than the loop
wedging.

Confirmed this is a real regression test, not a pseudo-test: reverted just
the `select!` arm head back to the old bare pattern and confirmed the test
fails (times out waiting for the echo) against that code, then confirmed
it passes again with the fix.

**Known gap:** `broker.rs`'s copy of this bug is fixed identically but
isn't covered by an equivalent automated test. Its `listen()` is a concrete
inherent method, not behind a generic `Service` impl like
`TcpListenerService`, so the same trait-abstraction test strategy would be
a substantially bigger refactor of the primary broker entrypoint. Given
this occurrence's severity is bounded (stalls at most until the next
maintenance/txn tick, not forever), that cost isn't justified here. This
half of the fix is verified by code review only, flagging it plainly rather
than claiming coverage we don't have.

```
cargo test -p nisshi-service --all-features   # 7 unit + 8 integration/doctest, all pass
cargo build -p nisshi-broker --all-features    # clean
cargo clippy -p nisshi-service -p nisshi-broker --all-features --all-targets -- -D warnings   # clean
cargo fmt --all --check                        # clean
```

## Out of scope (noted, not touched)

Other `tokio::select!` arms in the workspace use the same "bare pattern in
the arm head" shape and would silently disable on a mismatch the same way:
`nisshi-generator/src/lib.rs:361,381`, `nisshi-perf/src/lib.rs:491,511`
(`Ok(_) = ...`), and `nisshi-service/src/channel.rs:72`,
`nisshi-storage/src/service.rs:1295` (`Some(..) = rx.recv()`, disables on a
closed channel). Not fixed here, just flagging as follow-up candidates in
case they're useful to have on the list.


solace-aross:
Thanks for the thorough pass. Applied the cheap fixes: corrected the backoff doc comment (the sleep doesn't block resource-freeing, it delays cancellation/periodic branches), restored the listener address in the per-accept debug log, and moved `set_nodelay` above spinner creation so a failure there doesn't orphan it.

Left #1 (dedup broker.rs's backoff into the shared `AcceptBackoff`) and #5 (distinguish fatal listener errors) as follow-ups rather than folding them into this PR — both are real but neither blocks this fix, and #1 would mean threading `nisshi-service`'s type into `nisshi-broker`'s inherent method, a bit more churn than this PR's actual scope warrants.

shortishly:
I wonder if we should use something like: [tokio-retry](https://crates.io/crates/tokio-retry) for these kind of things?

Or, whether the whole listener should be a [TcpListener](https://docs.rs/rama/latest/rama/tcp/server/struct.TcpListener.html) instead?

PR #728, opened 2026-09-17
fix(broker): reject empty JoinGroup protocols instead of panicking

solace-aross:
Found this while going through the group-coordination path on a robustness pass, wanted to get it fixed quickly since it's a one-request crash.

## Problem

A `JoinGroupRequest` with a present-but-empty `protocols` array is a
legal, distinct wire encoding (the field has no `nullableVersions`, so
"absent" and "empty" are different states a client can send). For a
fresh group (no protocol negotiated yet), the existing guard only
caught the absent case:

```rust
let Some(protocols) = protocols else { /* return InvalidRequest */ };
```

An empty-but-present slice fell through to code that indexes
`protocols[0]` unconditionally, panicking the connection's handling
task. Any client can trigger this with one crafted request against any
group that hasn't negotiated a protocol yet.

## Fix

Added a guard rejecting `protocol_type.is_empty() || protocols.is_empty()`
before that indexing, returning `InconsistentGroupProtocol`. That's the
same code the negotiated-group path a few lines above already returns
for the equivalent condition (no matching protocol found), so this
keeps the function consistent with itself rather than returning
`InvalidRequest` for what's really the same case.

## Testing

New test `join_fresh_group_with_empty_protocols_is_inconsistent_group_protocol`,
through the real `Controller::join` entrypoint, asserting
`InconsistentGroupProtocol` and no panic. This one earned its keep as a
real test: my first assertion guessed the wrong `generation_id` and the
test failed against the actual (correct) implementation, so I fixed the
test, not the code.

```
cargo test --workspace --all-features join_fresh_group_with_empty_protocols   # 1 passed
cargo clippy -p nisshi-broker --all-features --all-targets -- -D warnings     # clean
cargo fmt --all --check                                                       # clean
```

(Used plain `cargo test` since `cargo-nextest` wasn't available in this
environment — worth a re-run under nextest before merge per the repo's
usual convention.)


solace-aross:
Good catch, thanks. Fixed: `protocol_name` is only nullable from v7+, and it's always `None` on this fresh-group path, so the branch was emitting a truncated frame for JoinGroup v0-v6. Changed it to `Some("".into())` to match every other error path in this function, and updated the test's expectation to match. Verified the test still passes and clippy/fmt are clean.

Also worth calling out plainly since it wasn't in the original description: this change also now rejects a fresh group joined with a non-empty `protocols` but empty `protocol_type`, which `main` previously accepted. That matches upstream Kafka's own `supportsProtocols` check, so it's correct, just flagging it as a real (and desirable) behavior change rather than a side effect nobody noticed.

Separately, noted but not fixing here since it's pre-existing and not part of this bug: a rejected fresh-group join is held ~22.5s before the client sees the error, since `Controller::join` sleeps once per iteration for any non-`MemberIdRequired` response. Same as the existing `InvalidRequest` path today, worth a follow-up on its own.

PR #729, opened 2026-09-17
fix(service): bound idle connections and stalled transfers with timeouts

solace-aross:
This one closes a loop from #723's own thread: that PR left "a read-idle timeout" as a noted follow-up when it merged. Same area, picking that up now.

## Problem

Neither `TcpBytesService::wait()` (reads the 4-byte length prefix) nor
`read()` (reads the declared frame body) nor `write()` (writes the
response) had any timeout. A peer that opens a connection and sends
nothing holds it, and its spawned task, open indefinitely,
pre-authentication. Separately, `read()` pre-allocated the *full*
declared frame size (up to 100 MiB) before a single body byte arrived,
so a peer that declares a large frame and trickles it slowly pins that
much memory immediately regardless of any timeout.

## Fix

Two-tier timeout on `TcpContext`, deliberately not one flat deadline
over the whole read/write (that would false-positive on a legitimate
client moving up to 100 MiB over a slow network):

- `connection_idle_timeout` (default 600s, matching Kafka's
  `connections.max.idle.ms`): how long an otherwise-idle connection may
  wait for the *next* request to start. Long on purpose, a real
  consumer/producer connection idling between requests is normal.
- `io_idle_timeout` (default 30s): how long a transfer already in
  progress may stall between chunks. Short and separate from the
  above, since once a peer has committed to sending/receiving it
  shouldn't stall for minutes. Reset on every chunk of forward
  progress, however small, so a slow-but-alive transfer is never
  penalized, only a genuinely stalled one.

`read()` now grows its buffer as bytes actually arrive instead of
pre-allocating the declared size up front. Idle-timeout disconnects are
routine, not exceptional, so they're logged at `debug!`, not `error!`,
in both `stream.rs` and `broker.rs`.

**What this doesn't fix, said plainly rather than overclaiming:** this
bounds how long a stalled connection/transfer can be held and how much
memory one stalled transfer pins at a time, but not how many
connections a peer can open at once. Someone opening many connections
and trickling each one just under the timeout still holds `N × ~64 KiB`
and `N` tasks/fds indefinitely. A max-connections cap is the
complementary fix (also noted in #723's follow-ups, alongside making
the 100 MiB cap configurable) and isn't part of this PR.

**Heads-up on overlap:** this touches the same "quiet error" match arm
in `broker.rs` as #727 (a different accept-loop fix), a few lines apart.
Whichever lands second will need a small, easy rebase.

## Testing

```
cargo check -p nisshi-service -p nisshi-broker --all-features --all-targets   # 0 errors, 0 warnings
cargo test -p nisshi-service --all-features                                   # 24 passed
cargo test -p nisshi-broker -p nisshi-storage -p nisshi-service --all-features
  # 27 passed; 7 pre-existing pg::* failures, all ConnectionRefused — no local
  # Postgres running in this environment, unrelated to this diff
cargo clippy -p nisshi-service -p nisshi-broker --all-features --all-targets -- -D warnings   # clean
cargo fmt --all --check                                                       # clean
```

New tests cover: a quiet connection closes after `connection_idle_timeout`;
a connection stalled mid-frame closes after `io_idle_timeout`; a
slow-but-steady transfer is not penalized.

Two notes for whoever re-runs this: `cargo-nextest` wasn't available in
this environment, so plain `cargo test` was used instead of this repo's
usual convention, worth a re-run under nextest before merge. Also,
`-p nisshi-broker --all-features` alone under-enables `nisshi-storage`
(its `dynostore`/`postgres`/`libsql` features don't forward the same-named
ones, only `slatedb` does) — pre-existing repo quirk, not touched here,
but worth knowing if a future check looks green for the wrong reason.


solace-aross:
This one's a fair catch, appreciate the injection test to prove it. Fixed the two tests so they actually pin which tier is which (disabled the other tier in each, added an outer 5s guard so a mis-tiered implementation fails fast instead of hanging). Re-verified the same way you did: swapped the tiers, confirmed both tests now fail immediately with a clear message, restored them, confirmed green.

Also corrected `DEFAULT_IO_IDLE_TIMEOUT`'s doc comment, it implied this timeout closes the "trickle one byte periodically" case, when it only bounds a genuine stall. Called that out as a tracked follow-up now rather than leaving the doc overclaiming it.

Left the other three as follow-ups rather than folding into this PR: the missing timeout-phase observability (agreed this matters, especially since the whole point is mitigating something that should be visible when it happens), the zero-fill overhead on the read buffer (correctness is fine, minor CPU cost only), and CLI config surface (matches this crate's existing precedent for `maximum_frame_size`, which also has no CLI wiring yet).

PR #730, opened 2026-09-17
fix(storage): bound Postgres pool acquisition and query execution

solace-aross:
Biggest of the four fixes from this pass, since it touches production Postgres behavior directly. Wanted extra care here given that.

## The problem

Nothing in the Postgres storage backend bounded how long a request could be stuck waiting on the database:

- The pool (`Pool::builder(mgr).max_size(16).build()`) had no `wait_timeout`/`create_timeout`, so `pool.get()` could block indefinitely once all 16 connections were checked out.
- No `statement_timeout` was set anywhere, so a single stalled query (e.g. a lock wait) held its connection open forever.

Once enough concurrent requests hit a stalled query, the pool exhausts and every other handler blocks forever in `pool.get().await`. A full, silent broker hang with no self-recovery.

## The fix

- **Pool-level timeouts.** `wait_timeout`/`create_timeout` (10s each) on the pool builder, plus `tokio_postgres::Config::connect_timeout` directly as defense in depth (note: in this crate version that only bounds the socket-level TCP connect per resolved address, not TLS/auth; deadpool's `create_timeout` is what bounds the full handshake). This required adding `.runtime(deadpool::Runtime::Tokio1)`: deadpool's `build()` returns `Err(BuildError::NoRuntimeSpecified)` once any timeout is configured without an explicit runtime. Missing this would have broken every `postgres://` broker's startup, so it's covered by a dedicated regression test (see Testing).
- **Server-side `statement_timeout` (30s default)**, set via the connection's startup `options` string rather than cancelling a stalled query client-side. Dropping a `tokio-postgres` query future client-side doesn't actually stop Postgres from executing it; the connection's background task keeps draining the response, and the connection stays busy until Postgres finishes, so it isn't safe to hand back to the pool early. A server-side timeout aborts the query cleanly and the connection is immediately reusable afterward. An operator's own explicit `statement_timeout` in their connection URL still wins over this default (the default is prepended, so their value, if present, is applied last per Postgres's `-c name=value` last-one-wins semantics).
- **Retriable error mapping.** A pool-acquisition timeout and a `statement_timeout` cancellation (SQLSTATE `query_canceled`) now map to `Error::Api(ErrorCode::RequestTimedOut)` instead of the previous generic `Error::Pool`/`Error::TokioPostgres` variants, which `produce.rs` was treating as non-retriable (`UnknownServerError`), turning a transient slowdown into a hard, non-retriable failure for the client. No changes needed in `produce.rs`; it already has an `Error::Api(_)` passthrough branch.
- **Maintenance-sweep exemption.** `policy_compact`/`policy_delete` run full-table-scan-shaped queries, invoked from a 10-minute maintenance tick. If the global 30s `statement_timeout` applied to them too, on any table where the sweep legitimately takes longer than 30s, it would get aborted every single tick, forever. The backlog only grows, it never gets a chance to catch up. Both now run `SET LOCAL statement_timeout = 0` as the first statement in their transaction, which reverts automatically at commit and doesn't affect any other query on that connection.

## Known, accepted trade-offs (revised after review, see below)

- **Real worst-case latency is higher than "10s + 30s."** `statement_timeout` is per statement, not per transaction, and a produce request runs several statements in one transaction. The honest bound is `wait_timeout + N x statement_timeout` for an N-statement transaction, which can meaningfully exceed a Kafka client's default 30s `request.timeout.ms`. The client will simply time out and retry independently; the broker still resolves the original request eventually with a retriable code, so this isn't a correctness problem, but the "40s" framing in an earlier draft of this PR understated it. Not engineering around this with a second, shorter request-path-specific timeout for now.
- **This changes a pre-existing hang into a livelock under sustained saturation, it doesn't remove the underlying cause.** `produce_in_tx` calls `describe_config`, which acquires a *second* pool connection while the transaction's own connection is still checked out. Under enough concurrent produces to schema-validated topics, every connection can end up waiting on one more that will never come free. Before this PR that was a silent, permanent hang. After it, those requests now fail with a retriable `RequestTimedOut` after `wait_timeout`, and clients retrying into the same saturated state can still livelock rather than recover. Strictly better (bounded and visible instead of silent and permanent), but the nested-acquisition root cause is still there. Filing as a follow-up rather than restructuring connection ownership in this PR.
- **This backend sets a `statement_timeout` via the connection's startup `options` string, which some connection poolers reject or silently drop.** PgBouncer only forwards a small fixed allowlist of startup parameters by default and errors on anything else, so fronting Postgres with PgBouncer as-is would break broker startup entirely; adding `ignore_startup_parameters = options` avoids the startup error but silently drops the parameter, meaning `statement_timeout` never actually applies with no signal that it didn't. Not verified against other poolers (RDS Proxy, Supavisor). If Nisshi needs to support a pooler in front of Postgres, the fix is a `post_create` hook that issues `SET statement_timeout` over the wire instead of a startup option, which works through a session-mode pooler the same way. Filing as a follow-up; flagging here rather than shipping silently.

## Testing

- `cargo test -p nisshi-storage --features postgres pg::tests`: **29 passed**, 0 failed.
- `cargo clippy -p nisshi-storage --all-features --all-targets -- -D warnings`: clean.
- `cargo fmt --all --check`: clean.
- New tests: 3 pure-function tests for the `statement_timeout` options-merge logic, 1 regression test that actually calls `Pool::builder(..).build()` and asserts `Ok` plus correctly-configured timeouts (specifically to catch the deadpool-runtime blocker above if it's ever reintroduced), 1 test that the `PoolError`/`SqlState::QUERY_CANCELED` to `RequestTimedOut` mapping is correct, and 2 integration-style tests exercising the `SET LOCAL statement_timeout = 0` exemption and the maintenance sweeps against a real Postgres.
- No local Postgres was reachable in the environment this was originally developed in, so those 2 integration tests skipped there. They have since been run for real against a live Postgres 17 (via review), all 29 tests passing against the live database, including a saturation probe confirming `connection()` now returns `Err(Api(RequestTimedOut))` after ~10s instead of hanging, and a direct check that a connection is safely reusable after a real `statement_timeout` cancellation.
- `cargo-nextest` wasn't available in the dev environment this was built in; ran via plain `cargo test` instead. Worth a nextest run before merge too.

## Deliberately out of scope

`nisshi-storage`'s libSQL backend (`lite.rs`) has a structurally similar pool-timeout gap (it also uses `deadpool`). Left untouched here, a separate, not-yet-triaged finding.


solace-aross:
Really appreciate the depth here, actually standing up a live Postgres and mutation-testing the runtime guard is well beyond what I could verify myself. Went through each finding:

- Fixed the `connect_timeout` doc comment (it overstated what that call bounds in this crate version).
- Rewrote the trade-off section: corrected the worst-case latency framing (it's `wait_timeout + N x statement_timeout` for a multi-statement transaction, not a flat 40s), and added honest notes on the nested-connection livelock risk and the PgBouncer startup-options gap, both filed as follow-ups rather than folded into this PR. Wanted those written down plainly rather than only living in this thread.
- Config surface for the three new constants: agreed this is fine for now, consistent with the pool size already being hardcoded.

Didn't attempt the `post_create`-hook alternative or the nested-acquisition restructuring here since both are real, separate pieces of work, not a one-line fix, and this PR's job was closing the hang, not redesigning connection ownership. Happy to pick either up as its own follow-up if that's the right call.

PR #734, opened 2026-09-18
fix(storage): keep the producer's codec and stop fetch looping per record on pg and lite

solace-aross:
Not for merging tonight. This is up for human review once the checks are confirmed green. It fixes the storage-engine gaps recorded in #733 and `compat/franz-go/FINDINGS.md`: on the `postgres` and `sqlite` legs of `compat-franz-go` and `compat-librdkafka`, `TestPooling` failed, `TestIssue865` ran to the 10 minute `go test` timeout, and librdkafka `0038_performance` consumed 0 of 83,886 messages. `memory` and `s3` were unaffected.

## Two root causes, both in the engines that rebuild batches from per-record rows

**1. Fetch looped once per record after `max_wait` (the hang).** `FetchService::fetch_partition` (`nisshi-storage/src/service/fetch.rs`) called storage until `max_bytes` was spent or nothing came back, with no deadline of its own. The pg and lite fetch paths stop assembling at the deadline but still return the record they were on, so once `max_wait` had passed every call returned exactly one record and re-ran the fetch query (a window sum over every remaining row) plus a header query. Ten thousand small records became thousands of round trips inside one Fetch response, longer than the client's request timeout. The client dropped the connection (the `BrokenPipe` every ~17s in the CI logs), reconnected, and re-asked the same offset, forever. Memory and S3 return whole batches per call, so the second call came back empty and the loop ended.

Fix: the loop returns what it has once `max_wait` has elapsed. While there, the next offset is now `max_offset() + 1` rather than `base_offset + record_count`, which under-advanced across compaction gaps and re-read rows already returned.

**2. Fetched batches lost the producer's compression.** Both engines stored `inflated.attributes` on each row, and `inflated::Batch::try_from(deflated::Batch)` clears the codec on purpose (the records are plaintext from there). So the batch rebuilt at fetch time was always sent uncompressed. Kafka with `compression.type=producer` (the default) hands back the producer's codec; memory and S3 already did, by storing the produced batch verbatim.

Fix: the row keeps the produced attributes (still validated through `BatchAttribute`, so reserved bits are masked as before) and fetch re-deflates with that codec. `limbo` gets the same one-line change. `nisshi-sans-io` gains the missing Snappy encoder, which is franz-go's default codec: without it, preserving the codec would have turned "uncompressed" into a fetch error. It writes a raw snappy block; franz-go, librdkafka, kafka-python, snappy-java and our own inflator all accept the raw form.

Not changed: `offset_stage` still returns a `Misuse` error for a just-deleted topic (item 3 in #733). It does not fail any test and the right fix is a per-partition error code in the fetch response, so it goes in its own PR.

## Tests

- `service::fetch::tests` in `nisshi-storage`: a scripted `Storage` that returns one record per call. Proves the loop stops at the deadline (10,001 storage calls before the fix) and advances past an offset gap (asked for offset 2 instead of 4 before the fix).
- `compression_preserved` in `nisshi-broker/tests/fetch.rs`, on every engine: one batch per codec (none, gzip, snappy, lz4, zstd), fetched back with the same codec and records. Red on lite with the produce change reverted (everything came back as one uncompressed batch); green on memory and slatedb with the produce change reverted, since they store the batch verbatim.
- `deflate_snappy_round_trip` in `nisshi-sans-io`; the previously `#[ignore]`d `produce_request_snappy` wire test is enabled.

## Verified locally

- franz-go v1.21.3 against `nisshi broker --storage-engine=sqlite://nisshi.db` built from this branch: `TestPooling` passes (failed `did not put decompress!` on the same binary before the fix); `TestIssue865` passes in 12s (timed out at 120s before). Full local suite results are in the first comment.
- `just fmt`, `just clippy`, `just check`, `cargo nextest` for `nisshi-sans-io`, `nisshi-storage` lib tests, and the `nisshi-broker` fetch tests on lite, memory and slatedb.
- Docker is not available on this machine, so postgres and the CI compat jobs are proven by this PR's own checks.

## Rollback note

Rows written after this change carry codec bits. A binary from before this change reads gzip/lz4/zstd rows but errors on fetch for any partition holding a snappy row (no encoder), and that applies to a field downgrade of a deployed broker as much as to a git revert. The safe path is roll forward, or keep the Snappy encoder arm when reverting. No schema change; rows written before this change keep behaving as they did.

## Follow-ups, not in this PR

- Per-record `header_fetch.sql` and the window sum over every remaining row bound the pg/lite fetch throughput; 0038 is now a throughput question rather than a hang.
- The first partition in a multi-partition Fetch consumes the whole `max_wait` budget; later partitions get one storage call with a zero budget (one record on pg/lite, one listed batch on dynostore).
- `s3://` with request batching enabled re-deflates combined batches uncompressed (`ProduceRequestBatcher::combine`); CI runs it without batching.

https://claude.ai/code/session_01MBdxBsm9zBWZXZHDXahMrf


solace-aross:
Local results against `nisshi broker --storage-engine=sqlite://nisshi.db` built from this branch (Docker is not available on the authoring machine, so postgres is proven by this PR's own checks).

**franz-go v1.21.3, full `compat/franz-go/run.sh` allowlist:** 55 of 55 reported tests pass (the remaining entries self-skip: SCRAM, share groups, KIP-848). `TestPooling` PASS, `TestIssue865` PASS in 11.8s. Before the fix, on the same machine: `TestPooling` FAIL (`did not put decompress!`), `TestIssue865` `panic: test timed out after 2m0s` in `PollRecords`.

**librdkafka v2.14.2, with `CI=1` (the 3.0 timeout multiplier the CI job runs with):**

A/B on the same machine, each binary on a fresh database:

| test | main (27a9be4) | this branch | note |
|---|---|---|---|
| 0038_performance | PASS, 30.0s, `consumed 0%` until 11.7s | PASS, 10.7s, no stall | the one failing test on the CI sqlite and postgres librdkafka legs, where the same initial stall runs into the 15s consume timeout |
| 0060_op_prio | PASS, 24.9s | PASS, 24.9s | |
| 0086_purge | PASS, 6.4s | PASS, 6.2s | |

0038 does not fail on this machine even before the fix, so its CI legs are the proof for that one. A first local pass of the full librdkafka allowlist without `CI=1` (1.0 multiplier) failed 0060 and 0086 on their 10s and 15s timeouts; both take the same time on main and on this branch and pass with the CI multiplier, so that was the local timeout setting, not this change. The remaining 72 tests passed on that pass.

**Workspace:** `cargo fmt --all --check`, `cargo clippy --workspace --all-features --all-targets -- -D warnings`, `cargo check --workspace --all-features --all-targets` clean. `nisshi-sans-io` 335 tests, `nisshi-storage` lib tests, `nisshi-broker` fetch tests on lite, memory and slatedb all pass.

PR #735, opened 2026-09-18
fix(broker): serve TLS on the listener when --cert and --key are given

sgamelin:
## Summary

`nisshi broker --cert X --key Y` accepted the flags but every connection stayed plaintext. Four independent defects lined up to make that silent; this PR fixes all four and adds the first tests that exercise `Broker::listen` end to end.

| Defect | Where | Fix |
|---|---|---|
| `TlsAcceptor` built then dropped, streams served raw | `nisshi-broker/src/broker.rs` | handshake runs inside each connection task, bounded by a 10s timeout; listener is TLS only when configured |
| `server_config(..).ok()` swallowed every cert/key error into plaintext | `nisshi-cli/src/cli/broker.rs` | errors fail startup and name the offending file; encrypted keys get a dedicated error |
| both flags in one clap `ArgGroup`, so together they conflicted and alone they short-circuited to `None` | `nisshi-cli/src/cli/broker.rs` | `requires` each other |
| `ring` and `aws-lc-rs` both compiled in, so `ServerConfig::builder()` panics picking a provider | `nisshi-cli/src/cli/broker.rs` | `builder_with_provider(aws_lc_rs)` explicitly, matching `nisshi-storage/src/pg.rs` |

`TcpContextService` in `nisshi-service` is now generic over the stream type so a `TlsStream` flows through the existing layer stack. Callers are unchanged. The `peer` span it used to carry moved to the spawn site in both `TcpListenerService` and the broker loop, at INFO, so error logs keep the client address.

### Behaviour change for existing deployments

- A broker already started with `--cert` and `--key` was serving plaintext. After this change the same command line serves TLS only; plaintext clients are refused at the handshake.
- `--cert` without `--key` (or the reverse) used to be accepted and ignored. It is now a clap error.
- Passphrase protected keys were never supported (rustls' PEM loader skips them); they now fail with a clear message instead of "no private key found".

The README `## broker` section documents all of this.

### Operability

- Startup: `TLS certificate <path> could not be loaded: ...` / `TLS private key <path> could not be loaded: ...` / passphrase and key-mismatch messages, printed by `nisshi` main instead of `Unknown error occurred`.
- Runtime: `info!("listener is tls only")` at bind so `--silent` deployments still show the mode in logs. Handshake failures from peers that connect and hang up (probes, scanners) log at debug; peers that spoke but could not negotiate log once at warn with the two likely causes.

## Tests

- `nisshi-broker/tests/tls.rs` (memory backend): plaintext round trip without TLS; TLS handshake plus `ApiVersions` and `Metadata` round trips; a plaintext client on a TLS listener is refused and the listener keeps serving TLS afterwards.
- `nisshi-cli` unit tests: flag pairing, valid config, missing/garbage/empty-chain/mismatched/encrypted (PKCS#8 and legacy `Proc-Type`) files, and `build()` failing on a bad cert before touching storage.

Before the fix these failed for the expected reasons (clap conflict, plaintext served, rustls provider panic). Verified locally: `just fmt`, `just clippy`, `typos`, `just test-doc`, and `just test-workspace` (876 tests; the 27 `nisshi-schema::berg` failures are pre-existing on `main` against my local minio and unrelated).

Manual: `openssl s_client` negotiates TLS 1.3 against a broker started with both flags; a raw `ApiVersions` frame gets a TLS alert back.

## Follow-ups (not in this PR)

- The bundled `nisshi cat`, `topic`, `perf` and `proxy` have no TLS client support, so they cannot talk to a TLS listener.
- A `TlsAcceptLayer` in `nisshi-service` would let the proxy share this instead of the broker's bespoke accept loop.
- A single shared crypto provider constructor (CLI, `nisshi-storage/src/pg.rs`, tests all hardcode `aws_lc_rs`).
- Handshake outcome metrics and a certificate expiry warning at startup.
- Encrypted private key support would need a PKCS#8 decryption dependency plus a passphrase source.


nisshi-broker/src/broker.rs
```diff
@@ -40,14 +40,42 @@ use tokio::{
...
+pub const TLS_HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// Report a failed TLS handshake at a level that matches its cause.
+///
+/// A peer that connects and hangs up (health checks, port scanners) is
+/// routine and logged at debug, mirroring how the plaintext path swallows
+/// EOF and reset. Anything else is a client that spoke but could not
+/// negotiate: almost always a plaintext client or one that does not trust
+/// the broker certificate, so the warning says so.
+fn handshake_failed(addr: SocketAddr, err: &std::io::Error) {
```
solace-wkourlas: I think these comments suffer greatly from poor AI writing, specifically that of Opus.

Perhaps the first comment could be something like:

```
/// Time limit for a client to complete the TLS handshake after its TCP
/// connection is accepted. Without it, a client that connects and never
/// sends a ClientHello would hold open rustls handshake state indefinitely.
pub const TLS_HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(10);
```

and for the second, the sentence that specifically reads poorly is: "Anything else is a client that spoke but could not negotiate: almost always a plaintext client or one that does not trust the broker certificate, so the warning says so."

I think in general function-level comments should be about function-level details, and implementation details of the function should be reserved to inside the function. So I think there doesn't have to be a function-level comment here, and instead a comment within the function stating "log at warn if the TLS negotiation failed in rustls and at debug if the TCP connection was closed or reset before the TLS handshake started" (or something).

solace-aross (approved):
All four defects are real and the fixes line up with root cause, not just symptom:

- The acceptor-built-then-dropped bug was the important one — `let _acceptor = ...` meant every connection was served raw regardless of `--cert`/`--key`. Verified this directly: reverted the fix locally (ignored `acceptor`, called `service.serve(c, stream)` unconditionally, reproducing the original code) and reran `nisshi-broker/tests/tls.rs`. `tls_handshake_and_kafka_request` and `plaintext_rejected_on_tls_listener` both failed exactly as expected — the plaintext client got back a real `ApiVersionsResponse` instead of a rejection. Restored the fix and reran: all three tests green, including the control (`plaintext_listener_without_tls`).
- `server_config(..).ok()` swallowing cert/key errors into a silent plaintext fallback, the `ArgGroup` conflict that made `--cert`/`--key` short-circuit to `None` when used together, and the dual `ring`/`aws-lc-rs` provider panic are all fixed at their actual source (explicit provider selection, `requires` instead of `group`, named errors instead of `.ok()`), not papered over.
- The handshake timeout (bounded at `TLS_HANDSHAKE_TIMEOUT`, inside the spawned task rather than the accept loop) is the right place for it: a slow/hostile peer can't stall new accepts, and the accept loop itself stays untouched.
- `TcpContextService` becoming generic over the stream type, with the peer span moving to the spawn site, is a clean way to let TLS streams flow through the existing `rama` layer stack without touching call sites — confirmed the `stream.rs` side of this (matches the same pattern in `broker.rs`).

Test coverage is real coverage, not just presence: `tls.rs` drives an actual `rustls` client through a real handshake and real `ApiVersions`/`Metadata` requests, and separately proves a rejected plaintext peer doesn't take the listener down with it. The `nisshi-cli` unit tests cover the operator-facing failure modes (missing/garbage/mismatched/encrypted key, `build()` failing before touching storage) with named, specific errors rather than a generic bail.

All 59 CI checks are green on this PR, and my own workspace build/test run confirms it independent of that.

The README's TLS section correctly calls out the behavior change for existing deployments (plaintext → refused) and the passphrase-key limitation that #737 addresses next.

solace-aross (approved):
Re-reviewed at the new head after the rebase onto main (rama 0.4 + storage crate split). This isn't a mechanical rebase, so I re-verified rather than trusting the old approval.

What I checked:

- Diffed `c3daa320`→`f541a4e6` scoped to the TLS PR's own files (`nisshi-broker/src/broker.rs`, `nisshi-broker/src/service.rs`, `nisshi-cli/src/cli/broker.rs`, `nisshi-broker/tests/tls.rs`, `nisshi-service/src/{api,frame,lib,stream}.rs`) rather than the full merge diff, which is mostly unrelated storage-crate-split churn from main.
- The TLS-validation-before-schema-registry ordering from the prior review still holds (`nisshi-cli/src/cli/broker.rs`: TLS config built at line 236, schema registry at 254+).
- `cargo check --workspace --all-features --all-targets` (matching `just check`) is clean, 0 errors.
- Ran the actual test suites rather than just reading them: `tls.rs` (3/3 pass — plaintext listener, plaintext-rejected-on-TLS-listener, full TLS handshake + Kafka ApiVersions/Metadata round trip) and `auth.rs` (6/6 pass, including `auth_handshake_scram_256_v1_reauth`, a multi-frame SASL handshake).

The reauth test mattered specifically: the merge changes how per-connection auth state propagates (`ctx.insert` + `assert!(...is_none())` under the old per-request `Context` → `extensions.contains::<AuthenticationExtension>()` guard under the new shared/interior-mutable `Extensions`). That's a real semantic shift, not just a rename — rama 0.4's `Extensions` appears to be a shared handle across a connection's frames rather than value-copied per clone, so the old assert would have fired spuriously on frame 2 of any connection under the new API if it had been kept as-is. The guard-and-skip rewrite is the correct adaptation, and the reauth test passing end to end is the proof, not just the type check.

`sasl_scram_enforcement.rs`'s `pg::scram_rejects_unverified_client` failed in my isolated worktree with `ConnectionRefused` — that's my worktree having no local Postgres (no `just ci` infra), not a code issue. The `in_memory`, `slatedb`, and `lite` backends of the identical logical test all pass.

No findings. Approving.

PR #737, opened 2026-09-21
feat(cli): accept passphrase-protected PKCS#8 keys via --key-passphrase-file

sgamelin:
Stacked on #735, which makes `--cert`/`--key` actually serve TLS but rejects every encrypted key. This PR adds `--key-passphrase-file <PATH>` so a passphrase-protected PKCS#8 key starts a TLS listener. It is the counterpart of Kafka's `ssl.key.password`.

## What it does

```shell
nisshi broker --cert broker.pem --key broker-key.pem --key-passphrase-file broker-key.passphrase
```

- The file holds the passphrase; trailing CR/LF bytes are ignored. `requires = "key"` in clap.
- Supported is what `openssl pkcs8 -topk8` emits: PKCS#8 `ENCRYPTED PRIVATE KEY`, PBES2 with PBKDF2-HMAC-SHA2 or scrypt, AES-CBC or Triple DES. Decryption happens once at startup via the `pkcs8` crate (features `encryption, pem, std, 3des`; no `sha1-insecure`, no `des-insecure`) and the plaintext DER is handed to rustls as `PrivateKeyDer::Pkcs8`.
- `--cert` and `--key` may point at one PEM bundle; the encrypted key section is located inside it.
- An empty passphrase file means no passphrase, so a mounted secret that is absent for unencrypted keys works unchanged. An encrypted key with an empty passphrase file still fails closed.
- A passphrase given for an unencrypted key is ignored with a `warn!`.
- TLS is still validated first in `build()`, before any registry, lake or storage connection, and there is no path back to a plaintext listener.

## Startup errors (all fatal, printed by `nisshi` main)

| Situation | Message |
|---|---|
| encrypted key, no flag | `TLS private key <path> is encrypted: pass --key-passphrase-file <file>.` |
| wrong passphrase | `TLS private key <path> could not be decrypted: ... Check the passphrase in --key-passphrase-file.` |
| SHA-1 PRF, PBES1, single DES | `TLS private key <path> is encrypted with PBKDF2 with HMAC-SHA1, which is not supported. ... Re-encrypt it: openssl pkcs8 -topk8 ... -v2 aes-256-cbc -v2prf hmacWithSHA256` |
| `Proc-Type: 4,ENCRYPTED` (legacy OpenSSL PEM) | `... uses legacy OpenSSL PEM encryption, which is not supported. Convert it: openssl pkcs8 -topk8 -in key.pem -out key-pkcs8.pem` |
| passphrase file unreadable | `TLS key passphrase file <path> could not be read: ...` |

The unsupported-algorithm case is split from the wrong-passphrase case on purpose: a raw OID plus "check the passphrase" would send an operator to retype a passphrase when the key needs re-encrypting.

## Not in scope

mTLS, certificate reload, expiry metrics, client-side TLS in `cat`/`topic`/`perf`/`proxy`, and legacy OpenSSL PEM decryption (single-round MD5 KDF; the conversion hint is the fix). Passphrase delivery is file-only, matching `--cert`/`--key`.

## Dependency note

`pkcs8` (RustCrypto) is a third crypto implementation next to `ring` and `aws-lc-rs`, used only for the one-off PBES2 decrypt at startup. `aws-lc-rs` has no PBES2/PKCS#8 parsing, so reusing it would mean hand-writing the ASN.1 and KDF plumbing. Flagged here so it is a conscious choice.

## Tests

`nisshi-cli` unit tests (31, all green): passphrase round trip via in-process PBES2 encryption, wrong/missing/empty passphrase, CRLF passphrase file, missing passphrase file, unencrypted key with passphrase, legacy PEM rejection, SHA-1 PRF rejection naming the algorithm, clap pairing, and `build()` end to end with an encrypted bundle on the memory backend.

OpenSSL interop fixtures under `nisshi-cli/tests/fixtures/tls/` (all generated locally, passphrase `correct-horse`, protecting nothing): EC P-256 (PBKDF2-SHA256 100k / AES-256-CBC), RSA-2048 with AES-256-CBC, Triple DES, scrypt, and the SHA-1 PRF negative. `nisshi-broker/tests/tls.rs` is untouched; the transport is unchanged.

Manual: `openssl s_client` negotiates TLS 1.3 against a broker started with the RSA cert+encrypted-key bundle and a passphrase file; wrong passphrase, no passphrase, legacy key and missing passphrase file each exit 1 with the line above; an unencrypted key with a passphrase file starts and logs the warning.

`just fmt`, `just clippy`, `typos` clean. The README `### TLS` section replaces #735's "passphrase protected keys are rejected" sentence with the flag, the PKCS#8-only rule, the conversion command, and a rollback note (earlier binaries do not know the flag).


solace-aross (approved):
Solid follow-on to #735. The layered error handling (legacy OpenSSL PEM check first, then PKCS#8 encrypted-section detection, then the plaintext fallback) is easy to follow and matches the PR description's table exactly, error message for error message.

All 31 `nisshi-cli` tests pass, and the coverage goes beyond what I expected: real `openssl pkcs8`-generated fixtures for PBKDF2-HMAC-SHA2 and scrypt, AES-CBC and 3DES, EC and RSA, plus the interop point that matters most — the in-process encrypt-then-decrypt round trip (`encrypted_key_with_passphrase_builds`) is checked *against* the OpenSSL fixtures rather than only against itself, so it can't mask a mismatch with what operators will actually produce.

Sabotaged the passphrase trim (`load_passphrase`) to strip only a single trailing `\n` instead of every trailing CR/LF, and reran: `encrypted_key_with_crlf_passphrase_builds` failed immediately with a decrypt error, exactly the class of bug (a mounted secret with a trailing CRLF from a Windows editor) the test exists to catch. Restored and confirmed green again.

Two things I checked because they're the kind of subtle bug that hides well in crypto code, both fine:
- The wrong-passphrase case: a bad passphrase can decrypt PBES2/AES-CBC to well-padded garbage without an error from `decrypt()`. The comment acknowledges this and defers to `with_single_cert`'s cert/key match, which is a real independent check, not hand-waving.
- The decrypted key material handed to rustls isn't zeroized (`PrivatePkcs8KeyDer` has no zeroize-on-drop), but that's identical to the existing unencrypted-key path from #735, not a regression this PR introduces, and the comment says so plainly instead of glossing over it.

All 59 CI checks green. No changes needed.

PR #740, opened 2026-09-22
feat(otel): honor standard OpenTelemetry resource environment variables for metrics

sgamelin:
Stacked on #737 (`feat/tls-key-passphrase`); review only the top seven commits.

## Problem

The OTLP metrics `Resource` was built with `Resource::builder_empty()` and a hardcoded `service.name`, in two separate copies (`nisshi-broker/src/otel.rs` for `nisshi broker`, `nisshi-otel/src/lib.rs` for `proxy` and `generator`). The SDK's `EnvResourceDetector` was never consulted, so the standard `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES` variables were ignored and operators could not attach `service.version`, `deployment.environment.name`, etc. to exported metrics.

## Behaviour

Every `key=value` pair in `OTEL_RESOURCE_ATTRIBUTES` is attached to the resource. Entries without a `=` are ignored with a warning, and values are taken as given. `service.name` is resolved with this precedence; unset, empty or whitespace-only values fall through to the next source:

| Priority | Source |
|---|---|
| 1 | `OTEL_SERVICE_NAME` (trimmed) |
| 2 | `service.name` inside `OTEL_RESOURCE_ATTRIBUTES` |
| 3 | Per-binary default: `nisshi-broker`, `nisshi-proxy`, `nisshi-generator` |

No `telemetry.sdk.*` attributes are added, so existing series are unchanged unless the operator sets these variables. The endpoint (credentials stripped), `service.name` and attribute keys are logged at `info` on startup.

`v1/metrics` is now appended to the endpoint path. Previously `Url::join` replaced the last segment, so `http://gateway/otlp` exported to `http://gateway/v1/metrics`.

## Changes

- **nisshi-otel**: new `resource(fallback_service_name)` builds the `Resource` from `EnvResourceDetector` plus the precedence above; new `metrics_endpoint()` appends `v1/metrics`; `meter_provider` uses both.
- **nisshi-broker**: depends on `nisshi-otel`, adds `Error::Otel`, and `metric_exporter` delegates to `nisshi_otel::meter_provider`. The duplicated exporter setup, the now-unused `Error::ExporterBuild` and the `opentelemetry-otlp` dependency are removed. The broker keeps the returned provider and force-flushes and shuts it down on exit, as proxy and generator already do, so the last export interval is not lost on restart.
- **nisshi-proxy / nisshi-generator**: no code change; they already call `nisshi_otel::meter_provider`.
- **README**: new `### Metrics` section under `## broker` covering the endpoint rule, `OTEL_EXPORTER_OTLP_HEADERS` for credentials, the resource variables, and an upgrade note.

## Upgrade note

Before this change `service.name` was always `nisshi-broker`. A deployment that already has `OTEL_SERVICE_NAME` or `OTEL_RESOURCE_ATTRIBUTES` in the broker's environment, for example injected for another agent, will see the broker's series move to that name. Unset the inherited variable or set `OTEL_SERVICE_NAME=nisshi-broker`.

## Tests

Eleven env-driven unit tests in `nisshi-otel` (via `temp-env`) cover the precedence cases, trimming, whitespace-only and empty values, malformed attribute entries, and a table of endpoint shapes (bare host, trailing slash, path prefix with and without slash). The first commits were red, then green.

Verified manually against Prometheus with `--web.enable-otlp-receiver`: with `OTEL_SERVICE_NAME=foo OTEL_RESOURCE_ATTRIBUTES=service.version=x,deployment.environment.name=dev`, `target_info` shows `job="foo"`, `service_version="x"`, `deployment_environment_name="dev"`. With the variables unset, series carry `service_name="nisshi-broker"`. `nisshi proxy` exports under `nisshi-proxy` through the same path. Also verified end to end in a Kubernetes deployment exporting through an OTLP collector.

Logs and traces are unchanged (they build no `Resource` today).


solace-aross (approved):
Approve. `nisshi-otel::resource()`'s hand-rolled precedence (`OTEL_SERVICE_NAME` > `service.name` in `OTEL_RESOURCE_ATTRIBUTES` > per-binary fallback) is the right call over `Resource::builder()`'s merge, which would let an empty `OTEL_SERVICE_NAME` silently win. Verified by running, not just reading: `cargo test -p nisshi-otel` (11/11 green), then sabotaged the precedence by swapping which source is checked first — `otel_service_name_wins_over_resource_attributes` went red exactly as it should, 10/11 stayed green. `cargo check --workspace --all-features --all-targets` and `cargo clippy -p nisshi-otel -p nisshi-broker --all-features --all-targets -- -D warnings` are both clean.

The broker's shutdown handling in `broker.rs` deliberately diverges from proxy's (`nisshi-proxy/src/lib.rs:357-365`, which propagates a flush/shutdown error with `?`): the broker warns and still returns `Ok(ErrorCode::None)`. That matches the PR's own framing ("a failed final export should not turn a clean shutdown into an error exit") and is arguably the better behavior of the two — flagging only so it's clear the divergence from proxy is intentional, not a miss.

Checked against the known cardinality open question on Nisshi's metrics (per-topic attributes on ~98 histogram call sites): resource attributes are process-level, attached once per export, not per-metric-point, so this doesn't interact with or worsen that risk.

No blocking findings.

PR #741, opened 2026-09-22
Add optional bearer token for the Iceberg REST catalog

csmithington:
## What

Adds an optional `ICEBERG_CATALOG_TOKEN` environment variable. When set, it is passed to the Iceberg REST catalog as the `token` property, which `iceberg-catalog-rest` uses for the `Authorization: Bearer` header on catalog requests.

When the variable is absent or empty, no `token` property is set and behavior is unchanged: local or unprotected catalogs continue to be contacted without authentication, exactly as before.

The value stays inside the catalog property map. It is not added to builder structs, request URLs, tracing output, or error messages, and it remains separate from the S3 access key pair.

Without this change, a bearer-token-protected catalog such as Cloudflare R2 Data Catalog rejects every request with `401 Unauthorized` from `iceberg_catalog_rest::catalog::RestCatalog::context`.

## Tests

Three focused tests in `nisshi-schema/src/lake/berg.rs`, no new dependencies:

- `rest_catalog_props_include_token_only_when_configured` — the `token` property is present only for a non-empty value.
- `rest_catalog_request_carries_bearer_token_when_configured` — drives the real `RestCatalogBuilder` against a loopback listener and asserts the first request carries `Authorization: Bearer`.
- `rest_catalog_request_omits_authorization_when_token_absent` — control asserting no `Authorization` header is sent when the variable is unset.

Verified on this branch: `cargo fmt --all --check`, `cargo check --workspace --all-features`, `cargo clippy --workspace --all-features --all-targets -- -D warnings`, and the focused catalog-token tests all pass.

## Qualification evidence

Pinned end-to-end run on a `v0.7.0-pre.2`-based build against a bearer-token-protected Cloudflare R2 Data Catalog: a KafkaJS client produced two synthetic archive topics, records survived a broker restart via broker-native read, and both DuckDB and Cloudflare R2 SQL returned the published rows with exact event IDs, sequence numbers, source timestamps, decimal text, and byte-identical payload JSON.


solace-aross (approved):
Clean, well-scoped change. `catalog_token()` only sets the property when non-empty, so unauthenticated catalogs are unaffected, and the token never reaches `debug!` or an error path.

Verified in a scratch worktree: ran the three new tests (all green), then sabotaged the fix (commented out the token insertion in `rest_catalog_props`) and confirmed `rest_catalog_props_include_token_only_when_configured` fails immediately with a clear assertion mismatch. The unit test alone catches a regression here without needing the network-backed tests.

One small gap worth a follow-up rather than blocking this: `ICEBERG_CATALOG_TOKEN` is read directly via `std::env::var`, bypassing the `clap` `env = ...` pattern every other broker config value uses (e.g. `ICEBERG_CATALOG`). That means it won't show up in `--help` or the README's flag table, and there's no mention of it in `example.env`. An operator setting up a bearer-token-protected catalog would only find this by reading the PR or the source. Worth a one-line README/`example.env` addition in a follow-up.

